PR #140, opened 2024-10-23
011 oct 19 getting input compliance tracker

melpsh:
## Release Notes

- **New Features**
  - Introduced `ComplianceProvider` for enhanced compliance state management across the application.
  - Updated `DropDowns` and `DatePicker` components to utilize compliance context, streamlining state management.
  - Added new save functionality in the `CustomModal` to log and format compliance data.

- **Bug Fixes**
  - Minor formatting adjustments in event handlers for improved code clarity.

- **Documentation**
  - Added new interfaces and context setup for compliance management, improving code structure and accessibility.

### DIFF
--- a/Clients/src/App.tsx
+++ b/Clients/src/App.tsx
@@ -21,37 +21,40 @@ import SetNewPassword from "./presentation/pages/Authentication/SetNewPassword";
 import ResetPasswordContinue from "./presentation/pages/Authentication/ResetPasswordContinue";
 import ProjectView from "./presentation/pages/ProjectView";
 import Playground from "./presentation/pages";
+import { ComplianceProvider } from "./provider/ComplianceContext";
 
 function App() {
   const mode = useSelector((state: any) => state.ui?.mode || "light");
   return (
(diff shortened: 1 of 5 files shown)


MuhammadKhalilzadeh (changes requested):
Hi
Thank you for the code 🙏 
Changes you need to apply as the first step:
No modifications specifically as a provider or wrapper shall be added to the App.tsx file, as it will be file specific but App.tsx is the root for the whole project.
Please find and use other methods

melpsh:
Hi Mohammad,

I took a look at your implementation of the Vendor component, you're managing states locally, however, in the DropDowns component, multiple inputs (like Select and DatePicker) share the state via Context since the state is also accessed in other parts of the application (like ControlPane). I use Context to avoid prop drilling and ensure that the state is easily accessible across multiple components without passing props through multiple layers. This simplifies maintenance, especially as the application grows and more components need access to the same state. For my component, since the state needs to be shared across different sections, using Context seems necessary to prevent redundancy and keep the code clean.
@MuhammadKhalilzadeh

PR #331, opened 2024-12-10
Move validateForm and passwordChecks to separate file(validations/formValidation)

samuel-coutinho:
The function `validateForm` has been duplicated across several files. To enhance code reusability, we can move it to a separate file and call it whenever it's needed.

Additionally, `passwordChecks` is currently being synchronized using `useEffect`. Since this approach is no longer recommended for syncing states in React (https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state), we could create a separate method for this purpose so that it can also be reused.

## Summary by CodeRabbit

- **New Features**
	- Introduced comprehensive form validation for user input, including password checks.
	- Added structured validation interfaces for form values and errors.

- **Bug Fixes**
	- Improved error handling during form submission based on new validation logic.

- **Refactor**
	- Streamlined validation logic in the `RegisterAdmin`, `RegisterUser`, and `SetNewPassword` components by removing unnecessary state management and using validation functions directly.
	- Updated the `validateForm` function to accept input values and return validation results instead of managing internal state.

- **Chores**
	- Removed unused imports related to previous validation methods.

### DIFF
--- a/Clients/src/application/validations/formValidation.ts
+++ b/Clients/src/application/validations/formValidation.ts
@@ -1,0 +1,123 @@
+import { checkStringValidation } from "./stringValidation";
+
+// Constants
+const PASSWORD_REGEX = /[!@#$%^&*(),.?":{}|<>]/;
+// Validation constants
+const VALIDATION_RULES = {
+  NAME: { min: 3, max: 50 },
+  PASSWORD: { min: 8, max: 16 },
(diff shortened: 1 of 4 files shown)


MuhammadKhalilzadeh (approved):
@samuel-coutinho 
Thank you so much for the implementation 🙏🏻

Issue #543, opened 2025-01-26
One time mock data insertion popup


MuhammadKhalilzadeh:
Here's the story:

Once the user creates his/her admin account for the first time, after login, a popup should be shown to ask if they want to proceed with mock data being inserted, or proceeding with an empty dashboard.

gorkem-bwl:
This is the [link to the design](https://www.figma.com/design/o4xu4PeC5bo1Ii4dyom6vQ/VerifyWise?node-id=3861-3263&t=c6Oxj7N7GYibiyC9-1). 

Here is the sample screenshot: 

<img width="683" alt="Image" src="https://github.com/user-attachments/assets/22799718-1360-400d-ac1b-f7704a2909b4" />

Process: 

- When the user clicks on Cancel, nothing happens. The user will be presented with an empty screen - the only user will be the admin. No projects will be shown. 

- When the user clicks on "Create demo data", that blue button should be disabled. At this time, data should be inserted into the database, and when this process is over, the modal disappears and user is presented with the demo data.

MuhammadKhalilzadeh:
Here's the component already created for you @samuel-coutinho 

![Image](https://github.com/user-attachments/assets/a458adc6-1a81-42eb-b3dd-1b3b417e8130)

samuel-coutinho:
If I understand correctly, this screen should only appear the first time the admin logs in. Do we have this information in BE? The way I see it, we need to save this information (if it is the admin's first login) in BE and send it to the app every time the admin logs in. What do you think @MuhammadKhalilzadeh and @gorkem-bwl ?

gorkem-bwl:
> Here's the component already created for you [@samuel-coutinho](https://github.com/samuel-coutinho) Inside pages/index.tsx (Playground)
> 
> ![Image](https://github.com/user-attachments/assets/a458adc6-1a81-42eb-b3dd-1b3b417e8130)

Can we align the strings to the left?

gorkem-bwl:
> If I understand correctly, this screen should only appear the first time the admin logs in. Do we have this information in BE? The way I see it, we need to save this information (if it is the admin's first login) in BE and send it to the app every time the admin logs in. What do you think [@MuhammadKhalilzadeh](https://github.com/MuhammadKhalilzadeh) and [@gorkem-bwl](https://github.com/gorkem-bwl) ?

[...]

PR #580, opened 2025-01-30
083 Jan 30: Fix auto driver

HarshP4585:
## Describe your changes

- Fixed the autoDrivers to insert the records
- Updated the delete mechanism for autoDrivers to reset the `id` key everytime
- Updated the delete mechanism to not drop the table

CC: @samuel-coutinho 

## Issue number

- #543 

## Summary by CodeRabbit

- **Data Management**
	- Modified data insertion and deletion logic in auto driver
	- Adjusted conditions for inserting mock data
	- Preserved table structures during data deletion

- **Mock Data Updates**
	- Commented out Admin role in role mock data
	- Removed "Alice" user from active user list

- **Utility Enhancements**
	- Reset ID sequence after data deletion
	- Modified data existence checking mechanism

### DIFF
--- a/Servers/driver/autoDriver.driver.ts
+++ b/Servers/driver/autoDriver.driver.ts
@@ -504,8 +504,7 @@ export async function insertMockData() {
     if (!(await checkTableExists(tableName as string))) {
       await createTable(createString as string);
     }
-    // if (tableName === "users" || tableName === "roles") {
-    if (await checkDataExists(tableName) === 1) {
+    if (await checkDataExists(tableName) > 1) {
       continue
     }
     if (mockData.length !== 0) {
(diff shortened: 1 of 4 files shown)


Servers/driver/autoDriver.driver.ts
```diff
@@ -504,22 +504,20 @@ export async function insertMockData() {
...
     }
-    // }
   }
 }
 
 export async function deleteMockData() {
   for (let tableName of deleteQueryExecutionOrder) {
     await deleteExistingData(tableName)
-    await dropTable(tableName)
+    // await dropTable(tableName)
```
samuel-coutinho, line 521: Should stopping the drop on the table reset the IDs? I have concerns about whether resetting the IDs of a production database is a good idea; I need to research this further. Perhaps when the mock data is created, we could save the project_id that was just created and pass it on to the other entities that depend on the project.
Another concern I had is that once users are not linked to an administrator, what would happen if several administrators tried to create demo data? Would we end up with duplicate mock data users?
What do you think @HarshP4585 , @gorkem-bwl  and @MuhammadKhalilzadeh ?

Issue #1009, opened 2025-03-18
Improve File upload component


samuel-coutinho:
Make UploadComponent perform all requirements from https://github.com/bluewave-labs/verifywise/issues/882

PR #123, opened 2024-10-20
004 oct 13 project risks view

popovaevgeniya:
Implemented Project Risks View and Vendor Risks View tabs

## Release Notes

- **New Features**
	- Introduced a new `Risks` component to display various risk categories.
	- Added a `RisksView` component for summarizing project risks with a structured table layout.
	- Enhanced the `ProjectView` component to include tabs for project and vendor risks.

- **Improvements**
	- Updated the `BasicTable` to conditionally render icons based on availability.
	- Streamlined the `Overview` component by integrating the `Risks` component for better modularity.
	- Refined the CSS styles for table components, improving visual presentation and spacing.

- **Bug Fixes**
	- Adjusted styling for improved layout and responsiveness in various components.

PR #1010, opened 2025-03-18
Add file upload functionality and delete confirmation modal

samuel-coutinho:
## Describe your changes
- Implemented DeleteFileModal component for confirming file deletions.
- Created getStyles utility for consistent styling in FileUpload component.
- Developed createUppy function to configure Uppy for file uploads.
- Updated VWQuestion component to handle file uploads and deletions.
- Removed unnecessary console log from questions fetching logic.

## Issue number

Mention the issue number(s) this PR addresses (e.g., #123).

[#1009](https://github.com/bluewave-labs/verifywise/issues/1009)

Screenshot_20250318_160344

## Summary by CodeRabbit

- **New Features**
  - Introduced a confirmation dialog for file deletion.
  - Rolled out enhanced file uploading with improved validations for file types and sizes.
  - Added a new component for displaying and managing uploaded files.
  - Implemented a new `DeleteFileModal` for confirming file deletions.
  - Added a new `UppyDashboard` component for improved file upload experience.
  - Introduced a structured representation for file-related data with the new `FileData` interface.

- **Refactor**
  - Streamlined file management and state handling for a smoother, more intuitive experience.
  - Redesigned the file listing and removal workflow to better reflect current file status.
  - Updated the `UppyUploadFile` component to simplify file handling and improve structure.

- **Chores**
  - Removed extraneous debugging output for cleaner operation.


Clients/src/application/tools/createUppy.tsx
MuhammadKhalilzadeh, line 1: Thank you so much for the code @samuel-coutinho 
One thing that we need to do here;

We must not have a `.tsx` file inside Application Layer, since it is against the purpose of the architecture.
So, here are the things I'd suggest.

1. Moving this `.tsx` file to Presentation Layer where this format is more relevant.
2. Changing the file to a `.ts` file (My suggestion)

I suggest this one since this is a function!

Clients/src/application/tools/createUppy.tsx
```diff
@@ -0,0 +1,92 @@
...
+ * @example
+ * const uppy = createUppy({
+ *   meta: { userId: "12345" },
+ *   onChangeFiles: (files) => console.log(files),
+ *   allowedMetaFields: ["userId"],
+ *   routeUrl: "upload/files",
+ *   authToken
+ * });
+ */
+const createUppy = ({
```
MuhammadKhalilzadeh, line 43: Now, about this being a function, we can have it like this, as an **arrow function**, like most of the components in `React.ts` being implemented, but we have a return here, so we need to go with function structure.

```
function createUppy ({}) : something {
...
return ... ;
}
```

Clients/src/application/tools/createUppy.tsx
```diff
@@ -0,0 +1,92 @@
...
+    },
+    allowedMetaFields,
+  });
+
+  uppy.on("upload-success", (_, response) => {
+    if (onChangeFiles && response?.body?.data) {
+      const files: FileData[] = [];
+      const data = response.body.data as string[];
+
+      data.map((file) => {
```
MuhammadKhalilzadeh, line 74: I have a suggestion for this part, to make it more semantically appealing:
Since you aren’t using the returned array, replacing `.map` with `.forEach` would be better.

Clients/src/application/tools/createUppy.tsx
```diff
@@ -0,0 +1,92 @@
...
+      maxFileSize: 10000000,
+      maxNumberOfFiles: 5,
+      allowedFileTypes: ["application/pdf"],
+    },
+  });
+
+  uppy.use(XHRUpload, {
+    endpoint: `${ENV_VARs.URL}/${routeUrl}`,
+    headers: {
+      Authorization: `Bearer ${authToken}`,
```
MuhammadKhalilzadeh, line 64: Here's another thing I'd suggest we do here:
Currently, even if `authToken` is undefined, the header will be set as Bearer undefined. It’s safer to include the header only when an auth token is provided
```
const headers = authToken ? { Authorization: `Bearer ${authToken}` } : {};
```

Clients/src/application/tools/createUppy.tsx
```diff
@@ -0,0 +1,92 @@
...
+    },
+    allowedMetaFields,
+  });
+
+  uppy.on("upload-success", (_, response) => {
+    if (onChangeFiles && response?.body?.data) {
+      const files: FileData[] = [];
+      const data = response.body.data as string[];
```
MuhammadKhalilzadeh, line 72: Although you’re casting `response.body.data` as a `string[]`, it might be beneficial to include runtime checks to ensure that the data is actually in the expected format. This would help avoid runtime errors in cases where the API response differs.

MuhammadKhalilzadeh (approved):
I just ran the code and tested it.
UX for working with files is fantastic!
Thank you @samuel-coutinho 

Let's work on the comments and then go for the merging

Clients/src/application/tools/createUppy.tsx
samuel-coutinho: Sorry, but I don't understand why we would need a regular function here. Arrow functions can also return values, as is the case here. I don't see any difference in using one or the other option. I usually prefer arrow functions because they are more concise and easier to read. Do you see any specific reason why we would need a regular function in this case?

Clients/src/application/tools/createUppy.tsx
samuel-coutinho: Ok, the code reviewers gave me an option for this question and that about using map or forEach.

MuhammadKhalilzadeh (approved):

MuhammadKhalilzadeh:
@samuel-coutinho We've got a small conflict here it seems

samuel-coutinho:
Hi @MuhammadKhalilzadeh , I have synced this branch and made some small changes to the Dashboard to make it easier to style.

MuhammadKhalilzadeh (approved):

MuhammadKhalilzadeh:
Thank you @samuel-coutinho

Issue #1036, opened 2025-03-24
Change the database image in docker-compose.yml


samuel-coutinho:
Our current Docker Compose configuration does not specify a PostgreSQL version, which means that the latest version will be used by default. This could introduce issues in our application if the version changes without being tested first.

We should modify this file to use a specific PostgreSQL version. This way, we can control when we update the version after thoroughly testing it locally.

PR #1045, opened 2025-03-26
Set a fix version for postgres in compose

HarshP4585:
## Describe your changes

- Set a fixed version of postges `17.4` in docker-compose

### NOTE

While running the application, if you encounter following error

image

- Run following command for your environment (**🔴 The data in the DB will be erased**):
    - DEV: `docker-compose --env-file .env.dev down --volumes`
    - PROD: `docker-compose --env-file .env.prod down --volumes`

## Issue number

- #1036 

## Summary by CodeRabbit

- **Chores**
  - Updated the backend database service to a fixed, stable version to ensure improved reliability and consistent performance.


docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: '3.9'
 
 services:
   postgresdb:
-    image: postgres
+    image: postgres:17.4
```
samuel-coutinho, line 5: I did some research, and it seems that not all cloud services support version 17, as it is quite recent. I would suggest using version 16 as it is the latest supported on most clouds. What are your thoughts on this?

HarshP4585: @samuel-coutinho I run the application with version 16 get `FATAL:  database files are incompatible with server`, cause the database was initialized with version 17.

1. Remove and Reinitialize the database volume, which will remove all the data from the db and it will start fresh
2. Migrate the data, in which we'll need to backup existing data, delete the volume and start fresh db with version 16, dump the existing data in the db

samuel-coutinho: Since the app is currently offline, I don’t see any issues with proceeding with option 1. What are your thoughts on this, @MuhammadKhalilzadeh?

MuhammadKhalilzadeh: I'm on board with that approach guys. Since the app is offline, option 1, removing and reinitializing the database volume is the quickest way to resolve the version incompatibility without introducing extra migration complexity. Therefore, if there's any risk of needing the current data later, it might be worth taking a backup before wiping the volume. Otherwise, proceeding as suggested should work fine.

HarshP4585: Okay 👍 Thank you @MuhammadKhalilzadeh @samuel-coutinho

samuel-coutinho (approved):

MuhammadKhalilzadeh (approved):

PR #1073, opened 2025-04-02
Fix vendor create bug

HarshP4585:
## Describe your changes

- Set `order_no` to `null` if not passed
image
image

CC: @srwa-rsp 

## Issue number

- NA

### DIFF
--- a/Servers/utils/control.utils.ts
+++ b/Servers/utils/control.utils.ts
@@ -83,7 +83,7 @@ export const createNewControlQuery = async (
       replacements: {
         title: control.title,
         description: control.description,
-        order_no: control.order_no,
+        order_no: control.order_no || null,
         status: control.status,
         approver: control.approver,
         risk_review: control.risk_review,
--- a/Servers/utils/controlCategory.util.ts
+++ b/Servers/utils/controlCategory.util.ts
@@ -71,7 +71,7 @@ export const createControlCategoryQuery = async (
       replacements: {
         project_id: controlCategory.project_id,
         title: controlCategory.title,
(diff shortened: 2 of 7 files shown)


Servers/utils/control.utils.ts
```diff
@@ -83,7 +83,7 @@ export const createNewControlQuery = async (
       replacements: {
         title: control.title,
         description: control.description,
-        order_no: control.order_no,
+        order_no: control.order_no || null,
```
samuel-coutinho, line 86: This field is defined as an `integer` in the DB. This will likely result in an error. If this value is required, we can change the frontend to always send it. Or we can set a proper default value. What do you think? @HarshP4585 , @gorkem-bwl , @MuhammadKhalilzadeh

Servers/utils/controlCategory.util.ts
```diff
@@ -71,7 +71,7 @@ export const createControlCategoryQuery = async (
       replacements: {
         project_id: controlCategory.project_id,
         title: controlCategory.title,
-        order_no: controlCategory.order_no
+        order_no: controlCategory.order_no || null
```
samuel-coutinho, line 74: The same comment applies here. Does this hold true for the other changes as well?

samuel-coutinho (changes requested):
I have some concerns about this approach. Could you please check this?

Servers/utils/control.utils.ts
HarshP4585: Yes, we can add default values, but as per the current structure, we can set null values for both of the tables `controls` and `controlcategories`

image
image

PR #1082, opened 2025-04-03
Move logout functionality in separate file

### LINKED ISSUE #1057: App should redirect users to the login page when the token expires.
Currently, if we reload a page or try to reach one with an expired token, the app doesn't redirect to the login page.
We need to fix that.

Since we are using Axios to make the API calls, I suggest adding something like this to our customAxios, in the `API.interceptors.response.use()` function:

```
api.interceptors.response.use(
  (response) => response,
  (error) => {
    if (error.response?.status === 401) {
      store.dispatch(logout()); // Clear token in Redux
      localStorage.removeItem("token"); // Clear localStorage
      window.location.href = "/login"; // Redirect to login
    }
    return Promise.reject(error);
  }
);
```

But in our case, we are using Redux, and we have a function to log out in `Clients/src/presentation/components/Sidebar/index.tsx` line 153:
```
  /**
   * Handles logging out the user
   *
   */
  const logout = async () => {
    // Clear the authentication token by dispatching the logout action
    dispatch(clearAuthState());

    // Navigate to the login page
    navigate("/login");
  };
```
I suggest moving this function to a separate file and calling it in both SideBar and customAxios.

charurajput16:
## Describe your changes

Move logout functionality to separate file to reuse at different places

Fixes #1057 

## Summary by CodeRabbit

- **New Features**
  - Introduced a unified sign-out mechanism that clears the user session and redirects to the login page.
  
- **Refactor**
  - Updated key components to adopt the new sign-out process, replacing redundant implementations.
  - Simplified logout functionality by utilizing a reusable hook.
  - Refined naming and related interactions for clearer account deletion workflows.

### DIFF
--- a/Clients/src/application/hooks/useLogout.ts
+++ b/Clients/src/application/hooks/useLogout.ts
@@ -1,0 +1,29 @@
+import { useDispatch } from "react-redux";
+import { useNavigate } from "react-router";
+import { clearAuthState } from "../authentication/authSlice";
+
+/**
+ * Custom hook for handling user logout
+ * 
+ * @returns {Function} A function that handles the logout process
+ */
+const useLogout = () => {
+  const dispatch = useDispatch();
+  const navigate = useNavigate();
+
+  /**
+   * Handles logging out the user
+   * Clears the authentication state and navigates to the login page
+   */
+  const logout = async () => {
+    // Clear the authentication token by dispatching the logout action
+    dispatch(clearAuthState());
+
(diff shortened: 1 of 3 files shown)


eieimon (approved):

eieimon:
@charurajput16 Thank you for this enhancement. 
CC: @MuhammadKhalilzadeh, I have done reviewing this pr.

MuhammadKhalilzadeh (approved):

PR #1160, opened 2025-04-12
#1104 multiple projects in vendors

### LINKED ISSUE #1104: A vendor should be assignable to more than one project.
A vendor should be assignable to more than one project. We can use the multi select dropdown component here instead of single select one. 

Also CC @HarshP4585 

Image

srwa-rsp:
## Describe your changes

Add support for multiple projects in vendors

Fixes #1104 

https://github.com/user-attachments/assets/4525ca6b-cc1a-40a9-9273-643a922364e1

## Summary by CodeRabbit

- **New Features**
  - Enhanced vendor modal interface with an autocomplete component, enabling users to select multiple projects rather than a single project.
  - Improved the selection experience by dynamically displaying and managing multiple project choices within the vendor details panel.

### DIFF
--- a/Clients/src/presentation/components/Modals/NewVendor/index.tsx
+++ b/Clients/src/presentation/components/Modals/NewVendor/index.tsx
@@ -534,16 +534,6 @@ const AddNewVendor: React.FC<AddNewVendorProps> = ({
                   <Typography sx={{ fontSize: "13px" }}>
                     {option.name}
                   </Typography>
-                  <Typography
-                    sx={{
-                      fontSize: "11px",
-                      color: "rgb(157, 157, 157)",
-                      position: "absolute",
-                      right: "9px",
-                    }}
-                  >
-                    {option.name}
-                  </Typography>
                 </Box>
               );
             }}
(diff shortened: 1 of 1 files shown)


MuhammadKhalilzadeh (changes requested):
Thank you @srwa-rsp 
Here are a couple of things that I'd like to address as my observations during the first review:

1. I'm really not sure about the overall layout. Can't we have it as it was? the first three input be in the same line? similar to lower sections? I guess we'll resolve the vertical overflow bug here automatically if we do so.
2. In case the overflow still exists, we can set a maxHeight for the modal itself and make it scrollable inside to address the overflow.
3. Also, in the multi-selector, you'll see that there will be overflows if you select long options or more that two options

Screenshot 2025-04-13 215329
Screenshot 2025-04-13 215255

srwa-rsp:
Hi @MuhammadKhalilzadeh, thanks for the feedback!
I adjusted the layout because I noticed that when a user selects multiple projects, the small input area makes it hard to see everything at once, they’d have to scroll through the projects in that tiny space. I think this new setup gives more room to view and manage the selections comfortably. Let me know if you'd like to tweak it further or revert to the previous layout!
Screenshot 2025-04-13 153632

MuhammadKhalilzadeh:
> Hi @MuhammadKhalilzadeh, thanks for the feedback! I adjusted the layout because I noticed that when a user selects multiple projects, the small input area makes it hard to see everything at once, they’d have to scroll through the projects in that tiny space. I think this new setup gives more room to view and manage the selections comfortably. Let me know if you'd like to tweak it further or revert to the previous layout! Screenshot 2025-04-13 153632

Thank you @srwa-rsp 
Looks better now without the overflow of the Selector:
1. Let's make the bottom border Selector algin with the bottom border of the Website input (More like a textarea)
2. Decrease the overall height of the modal, so the vertical overflow is resolved and it's scrollable

MuhammadKhalilzadeh:
What do you think @gorkemcetin ?

PR #1229, opened 2025-04-24
fix: Update risk data after a vendor is deleted

### LINKED ISSUE #1127: **Vendor Deletion & Related Updates**
- [x] Confirm that deleting a vendor appropriately updates related risk records (e.g., setting the vendor field to null or a default value).
- [ ] Verify that risk records remain editable and clickable after vendor deletion.

srwa-rsp:
## Describe your changes

Refetch vendor risks upon vendor deletion

Fixes  #1127

https://github.com/user-attachments/assets/49eeeb60-7726-4fd6-b703-59893e0c827f

## Summary by CodeRabbit

- **Bug Fixes**
  - Ensured the risks list is refreshed automatically after deleting a vendor.
- **Refactor**
  - Centralized vendor risk data fetching and state management into a dedicated hook.
  - Updated risk table to receive vendor risks data directly via props.
  - Improved loading state handling and re-fetch capabilities for vendor risks.

### DIFF
--- a/Clients/src/presentation/pages/Vendors/index.tsx
+++ b/Clients/src/presentation/pages/Vendors/index.tsx
@@ -62,7 +62,6 @@ const Vendors = () => {
   const theme = useTheme();
   const [isOpen, setIsOpen] = useState(false);
   const [isVendorsLoading, setIsVendorsLoading] = useState(true);
-  const [isRisksLoading, setIsRisksLoading] = useState(true);
   const [isRiskModalOpen, setIsRiskModalOpen] = useState(false);
   const [isSubmitting, setIsSubmitting] = useState(false);
   const [value, setValue] = useState("1");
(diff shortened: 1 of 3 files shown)


Clients/src/presentation/pages/Vendors/index.tsx
```diff
@@ -217,6 +217,7 @@ const Vendors = () => {
           setAlert(null);
         }, 3000);
         await fetchVendors();
+        await  fetchRisks();
```
samuel-coutinho, line 220: I believe this is a good opportunity to improve the component. The `useVendorRisks` hook, which is currently being utilized here(line 78) to provide `vendorRisksSummary`, can fetch and provide all the necessary data. Therefore, we can eliminate the `fetchRisks` function located on line 152.

Additionally, we could modify this hook. Instead of using a state variable (refreshKey) to control when the API is called, we could follow the approach I described in our meeting: create a `fetchVendorsRisks` function within the hook, call it inside a `useEffect` that runs only once, and expose the function as `refetchVendorRisks`. This way, we can call `refetchVendorRisks` in the component whenever it's needed.

samuel-coutinho (changes requested):
Thank you, @srwa-rsp. I have a suggestion to improve the fetch logic a bit.

Clients/src/presentation/pages/Vendors/index.tsx
srwa-rsp: Thanks for your feedback @samuel-coutinho ! I've made the changes as you suggested. Let me know if there’s anything else you’d like me to adjust.

MuhammadKhalilzadeh (approved):
Thank you @srwa-rsp 
It's working as explained
Kudos ✅

MuhammadKhalilzadeh (approved):

PR #1361, opened 2025-05-14
Add DELETE routes for ISO

HarshP4585:
## Describe your changes

- Added DELETE routes for ISO
   - `DELETE /iso-42001/:id`: Delete ISO framework for a project based on project id
   image

   - Updated `DELETE /project/:id` to delete iso framework when the project is deleted

Fixes #

### DIFF
--- a/Servers/controllers/iso42001.ctrl.ts
+++ b/Servers/controllers/iso42001.ctrl.ts
@@ -148,6 +148,9 @@ export async function deleteManagementSystemClauses(
   const transaction = await sequelize.transaction();
   try {
     const projectFrameworkId = parseInt(req.params.id);
+    if (!projectFrameworkId) {
+      return res.status(400).json(STATUS_CODE[400]("Invalid project framework ID"));
+    }
     const result = await deleteSubClausesISOByProjectIdQuery(projectFrameworkId, transaction);
 
     if (result) {
@@ -155,6 +158,7 @@ export async function deleteManagementSystemClauses(
       return res.status(200).json(STATUS_CODE[200](result));
     }
 
+    await transaction.rollback();
     return res.status(400).json(STATUS_CODE[400](result));
   } catch (error) {
     await transaction.rollback();
@@ -169,6 +173,9 @@ export async function deleteReferenceControls(
   const transaction = await sequelize.transaction();
   try {
     const projectFrameworkId = parseInt(req.params.id);
+    if (!projectFrameworkId) {
+      return res.status(400).json(STATUS_CODE[400]("Invalid project framework ID"));
+    }
     const result = await deleteAnnexCategoriesISOByProjectIdQuery(projectFrameworkId, transaction);
 
     if (result) {
@@ -176,6 +183,7 @@ export async function deleteReferenceControls(
       return res.status(200).json(STATUS_CODE[200](result));
     }
 
+    await transaction.rollback();
     return res.status(400).json(STATUS_CODE[400](result));
   } catch (error) {
     await transaction.rollback();
@@ -190,6 +198,9 @@ export async function deleteISO42001(
   const transaction = await sequelize.transaction();
   try {
     const projectId = parseInt(req.params.id);
+    if (!projectId) {
+      return res.status(400).json(STATUS_CODE[400]("Invalid project framework ID"));
+    }
     const result = await deleteProjectFrameworkISOQuery(projectId, transaction);
 
     if (result) {
@@ -197,6 +208,7 @@ export async function deleteISO42001(
       return res.status(200).json(STATUS_CODE[200](result));
     }
 
+    await transaction.rollback();
     return res.status(400).json(STATUS_CODE[400](result));
   } catch (error) {
     await transaction.rollback();
--- a/Servers/controllers/iso42001.ctrl.ts
+++ b/Servers/controllers/iso42001.ctrl.ts
@@ -1,10 +1,10 @@
-import { Response } from "express";
+import { Request, Response } from "express";
 import { sequelize } from "../database/db";
 import { SubClauseISO } from "../models/ISO-42001/subClauseISO.model";
 import { deleteFileById, uploadFile } from "../utils/fileUpload.utils";
 import { RequestWithFile, UploadedFile } from "../utils/question.utils";
 import { STATUS_CODE } from "../utils/statusCode.utils";
-import { updateAnnexCategoryQuery, updateSubClauseQuery } from "../utils/iso42001.utils";
+import { deleteAnnexCategoriesISOByProjectIdQuery, deleteProjectFrameworkISOQuery, deleteSubClausesISOByProjectIdQuery, updateAnnexCategoryQuery, updateSubClauseQuery } from "../utils/iso42001.utils";
 import { FileType } from "../models/file.model";
 import { AnnexCategoryISO } from "../models/ISO-42001/annexCategoryISO.model";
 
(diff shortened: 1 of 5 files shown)


Servers/utils/project.utils.ts
```diff
@@ -472,6 +474,8 @@ export const deleteProjectByIdQuery = async (
   for (let framework of frameworks) {
     if (framework.framework_id === 1) {
       await deleteProjectFrameworkEUQuery(id, transaction);
+    } else if (framework.framework_id === 2) {
+      await deleteProjectFrameworkISOQuery(id, transaction);
```
samuel-coutinho, line 478: We could consider using parallel execution and avoiding redundant `if/else` statements in this scenario. Instead of checking each framework ID individually, we can use a lookup object or a switch statement for better scalability. Additionally, running `deleteProjectFrameworkEUQuery` and `deleteProjectFrameworkISOQuery` concurrently with `Promise.all()` can enhance performance. Here’s an example:

```javascript
const frameworkDeletionMap: Record<number, (id: number, transaction: Transaction) => Promise<void>> = {
  1: deleteProjectFrameworkEUQuery,
  2: deleteProjectFrameworkISOQuery,
};

await Promise.all(
  frameworks.map(({ framework_id }) => {
    const deleteFunction = frameworkDeletionMap[framework_id];
    
    if (!deleteFunction) {
      throw new Error(`Unsupported framework_id encountered: ${framework_id}`);
    }

    return deleteFunction(id, activeTransaction);
  })
);
```

It’s also important to place this function in a reusable location, as we may need to use it in other parts of the code.

samuel-coutinho (changes requested):
Thanks, @HarshP4585 , I have a couple of suggestions.

Servers/controllers/iso42001.ctrl.ts
```diff
@@ -124,4 +124,67 @@ export async function saveAnnexes(
...
\ No newline at end of file
+}
+
+export async function deleteManagementSystemClauses(
+  req: Request,
+  res: Response
+): Promise<any> {
+  const transaction = await sequelize.transaction();
+  try {
+    const projectFrameworkId = parseInt(req.params.id);
```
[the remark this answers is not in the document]
HarshP4585: Added validation

Servers/controllers/iso42001.ctrl.ts
```diff
@@ -124,4 +124,67 @@ export async function saveAnnexes(
...
+    const result = await deleteSubClausesISOByProjectIdQuery(projectFrameworkId, transaction);
+
+    if (result) {
+      await transaction.commit();
+      return res.status(200).json(STATUS_CODE[200](result));
+    }
+
```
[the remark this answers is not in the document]
HarshP4585: Done

Servers/controllers/iso42001.ctrl.ts
```diff
@@ -124,4 +124,67 @@ export async function saveAnnexes(
...
+    return res.status(400).json(STATUS_CODE[400](result));
+  } catch (error) {
+    await transaction.rollback();
+    return res.status(500).json(STATUS_CODE[500]((error as Error).message));
+  }
+}
+
+export async function deleteReferenceControls(
```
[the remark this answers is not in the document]
HarshP4585: Done

Servers/controllers/iso42001.ctrl.ts
```diff
@@ -124,4 +124,67 @@ export async function saveAnnexes(
...
+    }
+
+    return res.status(400).json(STATUS_CODE[400](result));
+  } catch (error) {
+    await transaction.rollback();
+    return res.status(500).json(STATUS_CODE[500]((error as Error).message));
+  }
+}
+
+export async function deleteISO42001(
```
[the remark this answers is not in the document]
HarshP4585: Done

Servers/utils/project.utils.ts
HarshP4585: Updated the code

Servers/controllers/iso42001.ctrl.ts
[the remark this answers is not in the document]
samuel-coutinho: The check for `projectFrameworkId` being falsy (`if (!projectFrameworkId)`) will incorrectly fail for valid IDs like 0. Instead, we should use `isNaN(projectFrameworkId)` or a similar check to verify that the ID is a valid number. Additionally, this validation should be the first action taken, even before creating the transaction. If the request parameter is invalid, we should return early and avoid executing any further actions.

Servers/controllers/iso42001.ctrl.ts
```diff
@@ -169,13 +173,17 @@ export async function deleteReferenceControls(
   const transaction = await sequelize.transaction();
   try {
     const projectFrameworkId = parseInt(req.params.id);
+    if (!projectFrameworkId) {
+      return res.status(400).json(STATUS_CODE[400]("Invalid project framework ID"));
+    }
```
samuel-coutinho, line 178: The same comment for the previous function applies here.

Servers/controllers/iso42001.ctrl.ts
```diff
@@ -190,13 +198,17 @@ export async function deleteISO42001(
   const transaction = await sequelize.transaction();
   try {
     const projectId = parseInt(req.params.id);
+    if (!projectId) {
+      return res.status(400).json(STATUS_CODE[400]("Invalid project framework ID"));
+    }
```
samuel-coutinho, line 203: Here as well.

Servers/utils/project.utils.ts
```diff
@@ -471,13 +476,11 @@ export const deleteProjectByIdQuery = async (
...
-      await deleteProjectFrameworkEUQuery(id, transaction);
-    } else if (framework.framework_id === 2) {
-      await deleteProjectFrameworkISOQuery(id, transaction);
-    }
-  }
+  await Promise.all(
+    frameworks.map(async ({ framework_id }) => {
+      return await frameworkDeletionMap[framework_id](id, transaction);
+    })
+  );
```
samuel-coutinho, line 483: Thanks for the implementation, but I think the version that I suggested is better in this case:
```
await Promise.all(
  frameworks.map(({ framework_id }) => {
    const deleteFunction = frameworkDeletionMap[framework_id];
    
    if (!deleteFunction) {
      throw new Error(`Unsupported framework_id encountered: ${framework_id}`);
    }

    return deleteFunction(id, activeTransaction);
  })
);
```
It contains error checks for unknown or missing framework_ids and avoids unnecessary async/await in .map(). Let me know if you want to discuss it further.

samuel-coutinho (changes requested):
Thank you, @HarshP4585. I have a few additional comments. Please remember to resolve the conversations you have addressed.

Servers/controllers/iso42001.ctrl.ts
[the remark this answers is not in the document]
HarshP4585: The ids have SERIAL type and the value starts from 1, so, there won't be any instance with id 0

Servers/utils/project.utils.ts
HarshP4585: We are getting the framework ids from the db call itself before this query, we are not passing anything from the API. Do we want to add validation?

Servers/controllers/iso42001.ctrl.ts
HarshP4585: Added

Servers/controllers/iso42001.ctrl.ts
HarshP4585: Added

Servers/controllers/iso42001.ctrl.ts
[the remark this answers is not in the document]
HarshP4585: Added

Servers/controllers/iso42001.ctrl.ts
```diff
@@ -148,7 +148,7 @@ export async function deleteManagementSystemClauses(
   const transaction = await sequelize.transaction();
   try {
     const projectFrameworkId = parseInt(req.params.id);
-    if (!projectFrameworkId) {
+    if (isNaN(projectFrameworkId)) {
```
samuel-coutinho, line 151: Good catch with the `id` being more than 0. In this case, we want to validate for positive numbers, so we could use `isNaN(projectFrameworkId) || projectFrameworkId <= 0` inside the `if` condition.
Also, please place this validation before the transaction. Validating the input as early as possible is a fundamental principle of robust programming.  If the input is invalid, there's no need to begin a database transaction.  Transactions are relatively expensive operations, and we want to avoid starting them if we know the operation will fail due to bad input.
I will develop a better solution for validations soon, so that it is more reusable. For now, please do the same for the other 2 functions here.

samuel-coutinho (changes requested):

Servers/utils/project.utils.ts
HarshP4585: Updated the code

PR #1499, opened 2025-05-30
Fix "Generate report" disable

### LINKED ISSUE #1495: Fix GenerateReport isDisabled condition
Currently, the `GenerateReport` component disables the "Generate your report" button based on the existence of `selectedProjectId`. However, at this stage of the app, we do not have this value since it will be selected later.

```javascript
const { selectedProjectId } = dashboardValues;
const isDisabled = selectedProjectId ? false : true;
```

Instead, we should disable the button based on whether the length of the `projects` array is equal to 0.

Rather than using context to extract the projects, it would be more effective to call the `Clients/src/application/hooks/useFetchProjects.ts` hook to obtain the projects object.

Additionally, we should consider renaming the hook to `useProjects` to better reflect its purpose.

HarshP4585:
## Describe your changes

- Fix the "Generate report" to be disable if the length of project is more than 0
- Updated the project delete to update the projects in context
- Renamed the `useFetchProjects` to `useProjects`

https://github.com/user-attachments/assets/0d9effb9-82c6-4ae4-849c-fcc5a00890b5

Fixes #1495

## Summary by CodeRabbit

- **Refactor**
  - Renamed the project data hook for consistency across the application.
  - Updated all relevant components to use the new hook name for fetching project data.
  - Adjusted state initialization in report generation to assume project data is always present.
  - Improved project deletion handling to update the project list in real time after a project is removed.
  - Simplified logic for enabling the "Generate your report" button based on project availability.


Clients/src/presentation/pages/FileManager/index.tsx
```diff
@@ -11,7 +11,7 @@ import FileSteps from "./FileSteps";
 import VWSkeleton from "../../vw-v2-components/Skeletons";
 import { vwhomeHeading } from "../Home/1.0Home/style";
 import { useUserFilesMetaData } from "../../../application/hooks/useUserFilesMetaData";
-import { useProjectData } from "../../../application/hooks/useFetchProjects";
+import { useProjectData } from "../../../application/hooks/useProjects";
```
samuel-coutinho, line 14: We can rename the hook to `useProjects` as well to have the same name as the file.

samuel-coutinho (changes requested):

MuhammadKhalilzadeh:
@HarshP4585 You've got conflicts here
Please give me a heads up once you're done resolving

Clients/src/presentation/pages/FileManager/index.tsx
HarshP4585: Renamed the hook

HarshP4585:
@MuhammadKhalilzadeh resolved the conflicts

PR #1680, opened 2025-06-20
RiskChip component

charurajput16:
Use RiskChip component in table wherever severity or risk value is displayed.

## Describe your changes

Provide a concise description of the changes made and their intended purpose.

- [] I have included the issue # in the PR.
- [] The issue I am working on is assigned to me.

### DIFF
--- a/Clients/src/presentation/components/RiskLevel/RiskChip.tsx
+++ b/Clients/src/presentation/components/RiskLevel/RiskChip.tsx
@@ -1,0 +1,54 @@
+import { Chip, useTheme } from "@mui/material";
+import { useMemo } from "react";
+import React from "react";
+import { getSeverityColorByText, RISK_COLOR_BY_TEXT, getRiskChipStyle } from "./constants";
+
+interface RiskChipProps {
+  label: string;
+  type?: 'severity' | 'risk-level' | 'auto';
+}
+
+const RiskChip: React.FC<RiskChipProps> = React.memo(({ 
+  label, 
+  type = 'auto'
+}) => {
+  const theme = useTheme();
+  
+  // Memoize the backgroundColor calculation to avoid recalculation on every render
+  const backgroundColor = useMemo(() => {
+    if (!label) return '#B0B0B0';
+
+    if (type === 'severity') {
+      return getSeverityColorByText(label);
+    } else if (type === 'risk-level') {
+      return RISK_COLOR_BY_TEXT[label] || '#B0B0B0';
+    } else {
+      // Auto-detect: try severity first, then risk level
+      const severityColor = getSeverityColorByText(label);
+      if (severityColor !== '#B0B0B0') {
+        return severityColor;
+      }
+      return RISK_COLOR_BY_TEXT[label] || '#B0B0B0';
+    }
+  }, [label, type]);
+
+  // Memoize the chip style to avoid recreating the style object on every render
+  const chipStyle = useMemo(() => 
+    getRiskChipStyle(backgroundColor, theme), 
+    [backgroundColor, theme]
+  );
+
+  if (!label) {
+    return <span>-</span>;
+  }
+
+  return (
+    <Chip
+      label={label}
+      size="small"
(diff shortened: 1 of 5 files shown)


charurajput16:
Screenshot_2025-06-20_at_3 27 33_PM

Screenshot_2025-06-20_at_3 27 38_PM

Screenshot_2025-06-20_at_3 27 46_PM

gorkem-bwl:
LGTM!

Clients/src/presentation/components/RiskLevel/constants.ts
```diff
@@ -65,3 +65,50 @@ export const RISK_COLOR_BY_TEXT: Record<string, string> = Object.values(RISK_LAB
...
+
+// Reusable chip style object to avoid repetitive styling
+export const getRiskChipStyle = (backgroundColor: string, theme: any) => ({
+  backgroundColor,
+  color: 'white',
+  fontWeight: 500,
+  borderRadius: theme.shape.borderRadius,
+  height: 24,
+});
+
```
samuel-coutinho: Then here we could remove the backgroundColor.

Clients/src/presentation/components/RiskLevel/constants.ts
```diff
@@ -65,3 +65,50 @@ export const RISK_COLOR_BY_TEXT: Record<string, string> = Object.values(RISK_LAB
...
+      return RISK_LABELS.medium.color;
+    case 'high risk':
+      return RISK_LABELS.high.color;
+    case 'very high risk':
+      return RISK_LABELS.critical.color;
+    
+    default:
+      return '#B0B0B0';
+  }
+};
```
samuel-coutinho: Good implementation @charurajput16 . But I think that we could improve the logic a bit, something like this:
```
const SEVERITY_COLOR_LOOKUP: Record<string, string> = Object.entries(RISK_LABELS).reduce((acc, [key, val]) => {
  acc[key.toLowerCase()] = val.color;
  acc[val.text.toLowerCase()] = val.color;
  return acc;
}, {} as Record<string, string>);

export const getSeverityColorByText = (severity: string): string => {
  if (!severity) return '#B0B0B0';
  const color = SEVERITY_COLOR_LOOKUP[severity.toLowerCase().trim()];
  return color ?? '#B0B0B0';
};
```

This removes the need for the switch and makes future updates easier. And we can get rid of the `RISK_COLOR_BY_TEXT`

Clients/src/presentation/components/RiskLevel/RiskChip.tsx
```diff
@@ -0,0 +1,54 @@
...
+  return (
+    <Chip
+      label={label}
+      size="small"
+      sx={chipStyle}
+    />
+  );
+});
+
+export default RiskChip;
```
samuel-coutinho: We could avoid using the backgroundColor logic and repeating the logic by moving it outside this component:

```
import { Chip, useTheme } from "@mui/material";
import { useMemo } from "react";
import React from "react";
import { getSeverityColorByText, getRiskChipStyle } from "./constants";

interface RiskChipProps {
  label: string;
}

const RiskChip: React.FC<RiskChipProps> = React.memo(({ label }) => {
  const theme = useTheme();

  if (!label) {
    return <span>-</span>;
  }
  const normalizedLabel = label?.toLowerCase().trim() ?? '';

  const backgroundColor = useMemo(
    () => getSeverityColorByText(normalizedLabel),
    [normalizedLabel]
  );

  return (
    <Chip
      label={label}
      size="small"
      sx={{
      ...getRiskChipStyle(theme),
      backgroundColor,
  }}
    />
  );
});

export default RiskChip;
```

samuel-coutinho (changes requested):
Thanks @charurajput16 . I have a couple of suggestions. Let me know what you think.

MuhammadKhalilzadeh (approved):
@charurajput16 Looks very nice.
Thank you.
The code is good also, thanks

PR #1788, opened 2025-07-16
Fix organization update issue

HarshP4585:
## Describe your changes

- Updated the code to update the organization
   image
   image

NA

## Summary by CodeRabbit

* **Bug Fixes**
  * Disabled all organization member and project management features, including viewing, adding, and removing members or projects from organizations.
  * Organization details can now only be updated for name and logo fields.

* **Refactor**
  * Simplified the organization model and interfaces by removing all member, project, and demo-related properties and methods.
  * Organization endpoints and utilities related to members and projects are no longer available.


HarshP4585:
@samuel-coutinho The root cause for this issue was structure we defined the model here was different that the actual table we have implemented. The organization table didn't have any `members` and `projects` columns. So, when we are trying to get the object from ORM [here](https://github.com/bluewave-labs/verifywise/pull/1788/files#diff-1e302c480f4647facc8ed7d387ac8eb388331ff7b522dd68fede1358aa5e6a96L602) it looks for members and throws error.

Now, for the removed routes and controllers, we do not have functionality of adding any members to an organization (when we create a user he will already be a part of organization) or adding a project to a organization so they were idle. If this comes up in future, we can implement it, but as the current table structure does not allow this, it is better to remove in my opinion.

samuel-coutinho:
Thanks, @HarshP4585. Please go ahead and remove the unused code.

HarshP4585:
Removed @samuel-coutinho

PR #1956, opened 2025-08-22
Added Google Auth

HarshP4585:
## Describe your changes

- Added Google Auth


gorkem-bwl:
@MuhammadKhalilzadeh just fyi, we tested this yesterday with Harsh and there are a few minor things to do, but otherwise this should be ready.

Servers/database/migrations/20250821205021-update-users-table-for-google-auth.js
MuhammadKhalilzadeh, line 1: ### Top Areas for Improvement

1. **Constraint Logic Clarity and Robustness**
   - The check constraint users_password_null_if_google_id_present_check currently allows either google_id or password_hash to be non-null. However, the logic could unintentionally allow both to be null if there is a mistake elsewhere. If your intention is that at least one of them must be present, the constraint is correct. If you want to ensure that a user must have either a password or a Google ID (but not both), you may want to use XOR logic instead.
   - Consider making the logic explicit in comments for future maintainers.

2. **Migration Reversibility and Data Consistency**
   - The down migration sets password_hash back to NOT NULL, but if there are any users created via Google authentication who do not have a password_hash, this will fail. You may want to ensure that there are no rows violating this constraint before running the down migration, or add logic to handle those rows.

3. **Transaction Handling**
   - Use of transactions is excellent. However, running multiple DDL (ALTER TABLE) statements in a transaction may not be supported on all databases or could behave differently; this is usually safe in PostgreSQL, but it’s good practice to confirm compatibility or document any database requirements.

### Additional Notes

- The code is clear, uses async/await properly, and includes error handling with rollback, which is very good.
- Consider adding comments to explain the purpose of the migration and the business logic behind the constraints.

### Grade

**I would give this file a grade of A.**

- It is well-structured and follows best practices for migrations and error handling.
- The main point for improvement is handling data consistency on rollback and documenting constraint intent.

If you address the reversibility concern and add clarifying comments, this would be an A+ migration script.

Above, is an interesting review from GitHub Copilot @HarshP4585 . Let's apply the first two. 3rd one looks an interesting point as well, although our users will only use PostgreSQL I believe

MuhammadKhalilzadeh:
@HarshP4585 
Please give me a heads up once you're done applying changes

Servers/database/migrations/20250821205021-update-users-table-for-google-auth.js
HarshP4585: Hello @MuhammadKhalilzadeh as per you comments

1. The constraint `users_password_null_if_google_id_present_check` check if both the values should not be false and we are allowing both values to be present as well cause, a user can login with password or google. So, I think we are good here
2. In the down script, what should we store the database password for the users authenticated with google auth?
3. The comment says `his is usually safe in PostgreSQL` and we are using `PostgreSQL` so, here as well I think we are good

Can you please provide your inputs in **`#2`**

Thanks

MuhammadKhalilzadeh (approved):
Thank you @HarshP4585

Servers/database/migrations/20250821205021-update-users-table-for-google-auth.js
MuhammadKhalilzadeh: I believe we can make that field nullable, so user sets the password manually after registration and logging in.

MuhammadKhalilzadeh:
> @HarshP4585 Please give me a heads up once you're done applying changes

Let's do this in a second pr @HarshP4585 , as this one has already go too large.

PR #2035, opened 2025-09-04
Fix customAxios to not show token header in the URL

HarshP4585:
## Describe your changes

- Updated the customAxios to remove token header from the request params
   Before
   image
   After
   image

CC: @rachanabasnet 

### DIFF
--- a/Clients/src/application/hooks/useFrameworks.ts
+++ b/Clients/src/application/hooks/useFrameworks.ts
@@ -29,7 +29,7 @@ const useFrameworks = ({
   } = useQuery({
     queryKey: ['frameworks', 'all'],
     queryFn: async () => {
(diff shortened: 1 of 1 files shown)


Clients/src/infrastructure/api/customAxios.ts
```diff
@@ -80,6 +80,11 @@ CustomAxios.interceptors.request.use(
       config.headers.Authorization = `Bearer ${token}`;
     }
 
+    // Remove any authorization token from URL params or query string
```
rachanabasnet: I think rather than removing the auth token from URL query params here, we shouldn't pass the token as a param at all. Currently, it seems we are sending it as a query param in `apiServices.get` in `networkServices.ts`.

HarshP4585: Got it, updated

PR #2231, opened 2025-09-26
added-used-in-projects-dropdown-in-add-new-model

### LINKED ISSUE #2158: Add mapping from model inventory to projects or organizational frameworks
We need to add a mapping from projects to organizational frameworks. This can be done by adding "Mapped projects or organizational frameworks" line right under "Capabilities" and then adding a dropdown which includes all projects + all organizational frameworks (in our case, currently, ISO 42001 and ISO 27001). The drop down should be multi-select.

Image

swaleha456:
Add mapping from model inventory to projects or organizational frameworks
 
Fixes #2158 

Screenshot 2025-09-26 at 1 05 06 PM
Screenshot 2025-09-26 at 1 04 47 PM
Screenshot 2025-09-26 at 12 40 36 PM

### DIFF
--- a/Servers/utils/modelInventory.utils.ts
+++ b/Servers/utils/modelInventory.utils.ts
@@ -40,7 +40,7 @@ export const createNewModelInventoryQuery = async (
 
   try {
     const result = await sequelize.query(
-      `INSERT INTO "${tenant}".model_inventories (provider_model, provider, model, version, approver, capabilities, security_assessment, status, status_date, reference_link, biases, limitations, hosting_provider, is_demo, created_at, updated_at)       
+      `INSERT INTO "${tenant}".model_inventories (provider_model, provider, model, version, approver, capabilities, security_assessment, status, status_date, reference_link, biases, limitations, hosting_provider, used_in_projects, is_demo, created_at, updated_at)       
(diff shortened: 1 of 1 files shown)


Servers/domain.layer/interfaces/i.modelInventory.ts
```diff
@@ -15,6 +15,7 @@ export interface IModelInventory {
   biases?: string;
   limitations?: string;
   hosting_provider?: string;
+  used_in_projects: string;
```
HarshP4585, line 18: What is this field for? It is named like it should be a boolean field.

swaleha456:
> Thank you for the implementation @swaleha456
> 
> Only there are a few small warnings and errors that we need to fix.:
> 
> Image
> It's not letting me to create models

no worries @MuhammadKhalilzadeh I will check and fix then I will update here.

Servers/domain.layer/interfaces/i.modelInventory.ts
swaleha456: It’s new field for add new model in model Inventory to select the projects and frameworks are using in

swaleha456:
> > Thank you for the implementation @swaleha456
> > Only there are a few small warnings and errors that we need to fix.:
> > Image
> > It's not letting me to create models
> 
> no worries @MuhammadKhalilzadeh I will check and fix then I will update here.

It is fixed @MuhammadKhalilzadeh

MuhammadKhalilzadeh (approved):

PR #2335, opened 2025-10-07
Rb oct 02 slack automated notifications

### LINKED ISSUE #2257: Slack Notifications, Cron Jobs, Daily Notification Setup
For notifications like sending daily updates for `Policy Due Soon`, we need to set up cron jobs and Redis to run the notification in regular intervals and in the background. 

### Notification
- Policy due soon, still incomplete: *Policy due soon*· "{policy_name}" for {project_name} is due on {due_date} and is still incomplete Next steps: review evidence, assign editors, update status

### Routing rules
- Each notification type can send to many channels.
- If a type has no mapping, send to a default public channel.

rachanabasnet:
## Describe your changes

Added background jobs to process the policies that are due soon(7 days) and send Slack notifications to the user.
The job is queued every day at 9 am.

image

Fixes #2257 

- [X] I deployed the code locally.
- [X] I have performed a self-review of my code.
- [X] I have included the issue # in the PR.
- [X] I have labelled the PR correctly.
- [X] The issue I am working on is assigned to me.
- [X] I have avoided using hardcoded values to ensure scalability and maintain consistency across the application.
- [X] I have ensured that font sizes, color choices, and other UI elements are referenced from the theme.
- [X] My pull request is focused and addresses a single, specific feature.
- [X] If there are UI changes, I have attached a screenshot or video to this PR.

### DIFF
--- a/Servers/services/slack/policyDueSoonNotification.ts
+++ b/Servers/services/slack/policyDueSoonNotification.ts
@@ -50,8 +50,6 @@ export const sendPolicyDueSoonNotification = async (): Promise<number> => {
           await getAllPoliciesDueSoonQuery(tenantId);
 
         if (policies.length > 0) {
-          totalPoliciesProcessed += policies.length;
-
           // Send notification for each policy
           for (const policy of policies) {
             try {
@@ -89,6 +87,7 @@ export const sendPolicyDueSoonNotification = async (): Promise<number> => {
                     },
                     message,
                   );
+                  totalPoliciesProcessed += 1;
                   return userToNotify;
                 } catch (notificationError) {
                   // Log but don't stop processing other notifications


MuhammadKhalilzadeh:
@HarshP4585 and @Br0wnHammer would you please do a review here as soon as you get the time?

Servers/jobs/worker.ts
```diff
@@ -0,0 +1,24 @@
...
+
+const notificationWorker = createNotificationWorker();
+
+// Add workers here as you add on new workers within the application
+const workers: Worker[] = [notificationWorker];
+
+// Global error handler for all workers
+workers.forEach((worker) => {
+  worker.on("error", (err) => {
+    console.error(`Worker error on queue ${worker.name}:`, err);
```
Br0wnHammer: Instead of console logging the errors and the logs. You can use the logger we have. That way it is presented in a much cleaner way.

Servers/services/slack/policyDueSoonNotification.ts
```diff
@@ -0,0 +1,129 @@
...
+    for (const organization of organizations) {
+      const organizationId = organization.id!;
+      const tenantId = getTenantHash(organizationId);
+
+      try {
+        // Get all policies due soon for this tenant
+        const policies: PolicyManagerModel[] =
+          await getAllPoliciesDueSoonQuery(tenantId);
+
+        if (policies.length > 0) {
```
Br0wnHammer, line 52: Instead of adding all the policies in one go before they are processed. You should add it one by one as they are processed. In that case, if there are any discrepancy while processing the policy. It won't be added.

rachanabasnet: @Br0wnHammer How can we achieve this? We need policies that are due soon so the query will return all the policies that are due soon.

One thing we can do is query those policies individually before sending the notification, but in my opinion, that's just redundant since we're immediately querying the same thing again, just a few milliseconds later.

Servers/jobs/worker.ts
rachanabasnet: Updated!

MuhammadKhalilzadeh:
@rachanabasnet @HarshP4585 @Br0wnHammer 
Please give me a heads up once you're done with review, so all go for concluding and merging the PR.

Servers/services/slack/policyDueSoonNotification.ts
rachanabasnet: @Br0wnHammer Made changes based on feedback.

rachanabasnet:
@MuhammadKhalilzadeh @Br0wnHammer has completed the review.

PR #2398, opened 2025-10-11
Yb oct 11 fix task assignee duplicates

### LINKED ISSUE #2395: A person can be assigned more than once to a task
Screenshot: 

Image

Aryanak47:
## Describe your changes

 I fixed the validation issue where users could assign the same user multiple times in task assignment. Here are the changes made:

  - Frontend UI: Autocomplete with filterSelectedOptions prevents selection
  - Frontend Validation: Real-time and form submission validation
  - Backend Validation: Server-side validation prevents data corruption

Fixes #2395

image

image


Clients/src/presentation/components/Modals/CreateTask/index.tsx
MuhammadKhalilzadeh, line 1: Thank you @Aryanak47 
Let me add some analysis here using Copilot:

-------

1) How good they are?
- Positives:
  - Good UX intent: prevents/alerts duplicate assignees both in real-time and at submit time.
  - Showing an inline error message is helpful and accessible-looking (visually clear).
  - Removing filterSelectedOptions and explicitly filtering gives explicit control over shown options.
- Concerns:
  - The runtime guard around users is incomplete (see risk below).
  - The UX of silently preventing the update (early return) can be confusing unless the message is immediately visible/cleared consistently.
  - The error is set when duplicates are detected but not explicitly cleared when duplicates are resolved (no code shown to clear errors on successful selection).

2) How good the file is?
- Overall the changes improve validation and error feedback, which is good.
- However there are some robustness and UX issues that need attention (see improvements and breaking points). With fixes the changes will be solid and maintainable.

3) What can be improved?
- Prevent a potential runtime error when users is undefined:
  - Replace users?.map(...).filter(...) || [] with (users ?? []).map(...).filter(...). That avoids calling .filter on undefined.
    Example:
    const options = (users ?? [])
      .map(u => ({ id: u.id, name: u.name, surname: u.surname || "", email: u.email }))
      .filter(u => !values.assignees.some(a => a.id === u.id));
- Clear the assignees error when selection becomes valid:
  - After successful setValues call, clear errors.assignees:
    setErrors(prev => {
      const next = { ...prev };
      delete next.assignees;
      return next;
    });
- Consider not preventing the update silently:
  - Instead of returning and not updating values, you can either:
    - Remove duplicates automatically (dedupe newValue), or
    - Allow the update but show the error until submit (less surprising).
- Use consistent id typing and robust uniqueness check:
  - Ensure id types are consistent (strings vs numbers). Convert to string for Set if mixed types: const assigneeIds = newValue.map(a => String(a.id));
  - Or use a utility like lodash/uniqBy(newValue, 'id') for clarity.
- Memoize options:
  - Compute options via useMemo([users, values.assignees]) to avoid remapping on every render.
- Keep Autocomplete props consistent:
  - If you still want default Autocomplete behavior, re-evaluate removal of filterSelectedOptions (removing it is OK if you handle filtering manually, but ensure behavior matches intent).
- Accessibility:
  - Ensure the error Typography is linked to the input via aria-describedby or TextField/error props so screen readers announce it.
- Types & tests:
  - Add/confirm TypeScript types for assignee objects.
  - Add unit/interaction tests for duplicate handling and option filtering.

4) Can it cause any breaking?
Yes, there are potential breaking/runtime issues:
- Calling .filter after optional chaining:
  - If users is undefined, users?.map(...) returns undefined; calling .filter(...) on undefined will throw TypeError at runtime. Replace with (users ?? []).map(...).filter(...).
- UX/flow break:
  - Early return in handleAssigneesChange prevents values from updating when duplicates are chosen. If callers expect values.assignees to reflect the attempted change, this could break downstream logic.
- API/contract mismatch:
  - If backend or other parts of the app expect duplicates to be allowed (unlikely), blocking them could change behavior.
- Type mismatches:
  - If assignee.id types are inconsistent (number vs string), the Set-based duplicate detection might miss duplicates or treat distinct ids as equal unexpectedly.
- Removed filterSelectedOptions:
  - If other code relies on filterSelectedOptions being present for certain Autocomplete behavior, removing it might change selection/visibility behavior (though in this diff they implemented explicit filtering).
- Not clearing errors:
  - If errors.assignees is not cleared when duplicates are fixed, the UI could remain in a blocked/error state even after a user corrected the problem.

Concrete suggested edits (short)
- Safe mapping/filtering:
  const options = useMemo(() => {
    return (users ?? [])
      .map(user => ({ id: user.id, name: user.name, surname: user.surname ?? "", email: user.email }))
      .filter(u => !values.assignees?.some(a => a.id === u.id));
  }, [users, values.assignees]);

- Clear error on successful change:
  setValues(prev => ({ ...prev, assignees: newValue }));
  setErrors(prev => { const next = { ...prev }; delete next.assignees; return next; });

- Use stable duplicate check:
  const assigneeIds = newValue.map(a => String(a.id));
  const uniqueAssigneeIds = [...new Set(assigneeIds)];

Servers/utils/validations/tasksValidation.utils.ts
MuhammadKhalilzadeh, line 1: Adding a few more suggestions from Copilot here as well

--------

1) How good your changes are?
- Pros:
  - Simple and efficient for primitive values (strings/numbers): Set-based deduplication is O(n).
  - Clear error object with a code, field, and message—consistent with typical validation patterns.
  - Placed right before return so it contributes to the same errors array as other validations.
- Cons / caveats:
  - new Set(...) only de-duplicates by strict equality. If assignees are objects (e.g., { id: 'x', name: 'y' }) Set will treat distinct object references as different even if they represent the same user.
  - No normalization: '1' vs 1, or case differences in strings, will be considered distinct.
  - No type validation of each element (null, undefined, nested arrays, etc.) — could give surprising results.

2) How good the file is?
- From this diff we see consistent validation style (accumulating errors). That pattern is fine for express-style request validation.
- If the rest of the file follows the same structure, it's straightforward and easy to reason about.
- However, repeated patterns (like duplicate checks for other fields) should be factored into helper functions to improve DRY and testability.

3) What can be improved?
- Correctness for object assignees:
  - If assignees are objects, deduplicate by a stable key (e.g., id). Example keying logic: item => (typeof item === 'object' && item !== null ? item.id ?? JSON.stringify(item) : String(item)).
  - Or require/validate that assignees are primitive ids in the schema before this check.
- Robustness:
  - Validate element types (reject null/undefined) or coerce to string for comparison if desired.
  - Normalize strings (trim, lowerCase) if you want case-insensitive deduplication.
- API design / maintainability:
  - Extract duplicate-check logic into a reusable helper (e.g., validateNoDuplicates(arr, keyFn, fieldName, errorCode)) and reuse for creation/change validations (you noted “same as creation rules”).
  - Use constants/enums for the error code instead of a raw string literal.
  - Add unit tests covering primitive duplicates, object duplicates (by id), mixed types, and edge cases.
  - Consider using schema validation libraries (Joi / Zod) for clearer, central validation rules.
- Error message:
  - If clients depend on message text, consider i18n or consistent messages across validations.

Suggested improved implementation (concept)
- Example that deduplicates by id when items are objects, otherwise by stringified value:

const getAssigneeKey = (a: any) => {
  if (a && typeof a === 'object') return a.id ?? JSON.stringify(a);
  return String(a);
};

if (Array.isArray(data.assignees)) {
  const seen = new Set<string>();
  const hasDup = data.assignees.some(item => {
    const key = getAssigneeKey(item);
    if (seen.has(key)) return true;
    seen.add(key);
    return false;
  });
  if (hasDup) {
    errors.push({ field: 'assignees', message: 'Assignees cannot contain duplicates', code: 'DUPLICATE_ASSIGNEES' });
  }
}

4) Can it cause any breaking?
- Behavior change: yes — this introduces a new validation failure path. Clients that previously sent duplicate assignees (and relied on that) will now receive validation errors. That is a breaking behavioral change for any client/API consumer expecting duplicates to be allowed.
- Edge-case bugs (not strictly breaking but problematic):
  - If assignees are objects, duplicate detection may fail to detect duplicates unless keyed properly (leading to false negatives).
  - Type mismatches ('1' vs 1) may cause surprising behavior.
- Recommendation: treat this as a breaking change in the API contract. Communicate it in release notes or bump the API/patch accordingly, and add migration notes or server-side coercion if needed.

If you want, I can:
- Produce a small refactor patch that keys assignees by id (or by primitive value) and extracts a helper.
- Add suggested unit tests for the new validation. Which approach do you prefer?

Clients/src/presentation/components/Modals/CreateTask/index.tsx
Aryanak47: > Thank you @Aryanak47 Let me add some analysis here using Copilot:
> 
>     1. How good they are?
> 
> 
>     * Positives:
>       
>       * Good UX intent: prevents/alerts duplicate assignees both in real-time and at submit time.
>       * Showing an inline error message is helpful and accessible-looking (visually clear).
>       * Removing filterSelectedOptions and explicitly filtering gives explicit control over shown options.
> 
>     * Concerns:
>       
>       * The runtime guard around users is incomplete (see risk below).
>       * The UX of silently preventing the update (early return) can be confusing unless the message is immediately visible/cleared consistently.
>       * The error is set when duplicates are detected but not explicitly cleared when duplicates are resolved (no code shown to clear errors on successful selection).
> 
> 
>     2. How good the file is?
> 
> 
>     * Overall the changes improve validation and error feedback, which is good.
> 
>     * However there are some robustness and UX issues that need attention (see improvements and breaking points). With fixes the changes will be solid and maintainable.
> 
> 
>     3. What can be improved?
> 
> 
>     * Prevent a potential runtime error when users is undefined:
>       
>       * Replace users?.map(...).filter(...) || [] with (users ?? []).map(...).filter(...). That avoids calling .filter on undefined.
>         Example:
>         const options = (users ?? [])
>         .map(u => ({ id: u.id, name: u.name, surname: u.surname || "", email: u.email }))
>         .filter(u => !values.assignees.some(a => a.id === u.id));
> 
>     * Clear the assignees error when selection becomes valid:
>       
>       * After successful setValues call, clear errors.assignees:
>         setErrors(prev => {
>         const next = { ...prev };
>         delete next.assignees;
>         return next;
>         });
> 
>     * Consider not preventing the update silently:
>       
>       * Instead of returning and not updating values, you can either:
>         
>         * Remove duplicates automatically (dedupe newValue), or
>         * Allow the update but show the error until submit (less surprising).
> 
>     * Use consistent id typing and robust uniqueness check:
>       
>       * Ensure id types are consistent (strings vs numbers). Convert to string for Set if mixed types: const assigneeIds = newValue.map(a => String(a.id));
>       * Or use a utility like lodash/uniqBy(newValue, 'id') for clarity.
> 
>     * Memoize options:
>       
>       * Compute options via useMemo([users, values.assignees]) to avoid remapping on every render.
> 
>     * Keep Autocomplete props consistent:
>       
>       * If you still want default Autocomplete behavior, re-evaluate removal of filterSelectedOptions (removing it is OK if you handle filtering manually, but ensure behavior matches intent).
> 
>     * Accessibility:
>       
>       * Ensure the error Typography is linked to the input via aria-describedby or TextField/error props so screen readers announce it.
> 
>     * Types & tests:
>       
>       * Add/confirm TypeScript types for assignee objects.
>       * Add unit/interaction tests for duplicate handling and option filtering.
> 
> 
>     4. Can it cause any breaking?
>        Yes, there are potential breaking/runtime issues:
> 
> 
>     * Calling .filter after optional chaining:
>       
>       * If users is undefined, users?.map(...) returns undefined; calling .filter(...) on undefined will throw TypeError at runtime. Replace with (users ?? []).map(...).filter(...).
> 
>     * UX/flow break:
>       
>       * Early return in handleAssigneesChange prevents values from updating when duplicates are chosen. If callers expect values.assignees to reflect the attempted change, this could break downstream logic.
> 
>     * API/contract mismatch:
>       
>       * If backend or other parts of the app expect duplicates to be allowed (unlikely), blocking them could change behavior.
> 
>     * Type mismatches:
>       
>       * If assignee.id types are inconsistent (number vs string), the Set-based duplicate detection might miss duplicates or treat distinct ids as equal unexpectedly.
> 
>     * Removed filterSelectedOptions:
>       
>       * If other code relies on filterSelectedOptions being present for certain Autocomplete behavior, removing it might change selection/visibility behavior (though in this diff they implemented explicit filtering).
> 
>     * Not clearing errors:
>       
>       * If errors.assignees is not cleared when duplicates are fixed, the UI could remain in a blocked/error state even after a user corrected the problem.
> 
> 
> Concrete suggested edits (short)
> 
>     * Safe mapping/filtering:
>       const options = useMemo(() => {
>       return (users ?? [])
>       .map(user => ({ id: user.id, name: user.name, surname: user.surname ?? "", email: user.email }))
>       .filter(u => !values.assignees?.some(a => a.id === u.id));
>       }, [users, values.assignees]);
> 
>     * Clear error on successful change:
>       setValues(prev => ({ ...prev, assignees: newValue }));
>       setErrors(prev => { const next = { ...prev }; delete next.assignees; return next; });
> 
>     * Use stable duplicate check:
>       const assigneeIds = newValue.map(a => String(a.id));
>       const uniqueAssigneeIds = [...new Set(assigneeIds)];

Hi @MuhammadKhalilzadeh ,I was waiting for the review. After your review, I have already implemented the suggested feedback.

MuhammadKhalilzadeh: Thank you @Aryanak47

PR #2422, opened 2025-10-16
Yb oct 9 soft delete risks

### LINKED ISSUE #2190: Historical data and analytics for `Risks`: Part 1
For all the different types of the risks that we have (`projectRisks`, `vendorRisks`, etc.), from this point forward, the DELETE operation will no longer remove or wipe them out entirely from the platform.
We need to add another columns to their schemas as `is_deleted`. When the user attemps to delete, this field will simple turn `true`, and the get request will not show it in the normal case.

Now, in the frontend, anywhere that we are getting risks using `GET` request, we can use filters to:
Only show the existing (active) data
Show both existing and deleted data
Only show the deleted data.

Also, the following fields/columns must exist as well: `created_at`, `updated_at`, `deleted_at`; all of them as a `Date`.

Aryanak47:
## Describe your changes

Backend: 
Added another columns to different risk's schemas as is_deleted. When the user attempts to delete, this field will simple turn true, and the get request will not show it in the normal case.Also, modified controller,utils accordingly.

Fixes #2190


MuhammadKhalilzadeh:
@rachanabasnet @swaleha456 @solan117 
Would you please do a review here asap?

rachanabasnet (approved):

rachanabasnet:
@MuhammadKhalilzadeh LGTM!

PR #2773, opened 2025-11-28
Fix migration errors by adding IF NOT EXISTS

gorkem-bwl:
## Summary
Fix migration errors when tables already exist in the database.

## Changes
- Add `IF NOT EXISTS` to CREATE TABLE statements in:
  - `20251122165467-add-ce-marking-associations.js`
  - `20251122165512-add-ce-marking-incidents.js`

## Test plan
- [ ] Run migrations on a database where tables already exist
- [ ] Verify migrations complete without errors


MuhammadKhalilzadeh (approved):
LGTM and works without breaking

PR #2847, opened 2025-12-08
Fix dashboard greeting flickering on initial load

gorkem-bwl:
## Summary

Fixes the dashboard greeting message flickering rapidly between different late night messages (e.g., "Still up? You're dedicated" and "Night owl mode activated") on initial page load.

## Root cause

The `getTimeBasedGreeting` function used `Math.random()` to select a late night greeting message. This function was called via `useMemo` with `[userName, userToken]` as dependencies. During initial load, when user data updates from the API, the memo recalculates and picks a different random message each time - causing rapid flickering.

## Solution

Replaced `Math.random()` with a stable index based on current minute:
```javascript
const stableIndex = now.getMinutes() % lateNightMessages.length;
```

This ensures:
- Consistent message during re-renders (no flickering)
- Message still rotates every minute for variety

## Test plan

- [ ] Visit dashboard between 1:00 AM and 4:00 AM
- [ ] Verify greeting message does not flicker on load
- [ ] Verify greeting message changes if you wait a minute and refresh

### DIFF
--- a/Clients/src/presentation/pages/DashboardOverview/IntegratedDashboard.tsx
+++ b/Clients/src/presentation/pages/DashboardOverview/IntegratedDashboard.tsx
@@ -111,8 +111,10 @@ const getTimeBasedGreeting = (userName?: string, userToken?: any): { icon: React
     ];
 
     if (hour >= 1 && hour <= 4) {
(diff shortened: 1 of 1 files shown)


gorkem-bwl:
Check if this happens in other message types

PR #2919, opened 2025-12-17
LLM Evals Update

EfeAcar6431:
## Describe your changes

Backend - Evaluation Engine
Implemented actual model response generation for multi-turn datasets (model now generates responses, not just evaluates pre-written ones)
Added dedicated conversational metrics: Turn Relevancy, Knowledge Retention, Conversation Coherence, Conversation Helpfulness, Task Completion, Conversation Safety
Added per-turn Bias/Toxicity evaluation on each assistant response
Fixed Bias/Toxicity pass/fail logic (lower = better for safety metrics)
Created get_metrics_for_evaluation(use_case, is_multi_turn) as single decision point for all metric selection
Removed all fallback code for production readiness
Fixed gatekeeper to skip missing metrics instead of failing

Frontend - UI Updates
Added chat-style display for multi-turn conversations in experiment detail view
Updated NewExperimentModal to show appropriate metrics based on use case (Chatbot/RAG/Agent) and dataset type (single/multi-turn)
Improved multi-turn dataset editor with proper styling (icons, colors, alignment)
Added "Multi-turn dataset detected" indicator with conversational metrics accordion

Metrics Logic
Chatbot + Multi-turn → All conversational metrics
Chatbot + Single-turn → Universal core metrics
RAG/Agent → Base metrics + use-case specific metrics

Screenshot 2025-12-17 at 08 32 32
Screenshot 2025-12-17 at 08 32 16
Screenshot 2025-12-17 at 08 31 50

## Write your issue number after "Fixes "

Fixes #2918 

## Please ensure all items are checked off before requesting a review:

- [x] I deployed the code locally.
- [x] I have performed a self-review of my code.
- [x] I have included the issue # in the PR.
- [x] I have labelled the PR correctly.
- [x] The issue I am working on is assigned to me.
- [ ] I have avoided using hardcoded values to ensure scalability and maintain consistency across the application.
- [ ] I have ensured that font sizes, color choices, and other UI elements are referenced from the theme.
- [ ] My pull request is focused and addresses a single, specific feature.
- [ ] If there are UI changes, I have attached a screenshot or video to this PR.

PR #2920, opened 2025-12-17
Fix Mistral response handling when content is returned as list

sermengi:
## Describe your changes

When running experiments with **Mistral models**, experiment creation failed with the error: 
```
'list' object has no attribute 'strip'
``` 
This occurred because the Mistral SDK may return `message.content` as a **list of content blocks** instead of a plain string, which was not handled in the generation pipeline.

#### Solution
This PR updates the Mistral generation logic to:
- Detect when `message.content` is a list
- Extract and concatenate text from content blocks (e.g. `{ "text": ... }`)
- Safely normalize the output to a string before calling `.strip()`

This PR does not intend to fix any issue.


gorkem-bwl:
Lets wait for Efe's fixes (https://github.com/bluewave-labs/verifywise/pull/2919) first and then this one would go in next.

PR #2958, opened 2025-12-22
Feat: Azure Mail Tests

### LINKED ISSUE #2427: Testing email services (Azure)
We need to test the newly added email services

- [x] Exchange Online (Microsoft 365/Exchange Online) (@Br0wnHammer ) 
- [x] Amazon SES (Simple Email Service) (@Br0wnHammer) 

(SES works)

Br0wnHammer:
## Describe your changes

This PR implements azure communications tests.

Fixes #2427 

### DIFF
--- a/Servers/services/email/providers/AzureCommunicationServicesProvider.ts
+++ b/Servers/services/email/providers/AzureCommunicationServicesProvider.ts
@@ -1,0 +1,91 @@
+import { EmailClient, EmailMessage } from '@azure/communication-email';
+import { EmailProvider, EmailOptions, EmailResult } from '../types';
+
+/**
+ * Azure Communication Services Email Provider
+ * 
+ * Uses Azure Communication Services Email SDK with connection string authentication.
+ * Simple setup - just needs the connection string from Azure Portal.
+ */
+export class AzureCommunicationServicesProvider implements EmailProvider {
+  private client: EmailClient;
+  private connectionString: string;
+
+  constructor(connectionString: string) {
+    this.connectionString = connectionString;
+    this.client = new EmailClient(connectionString);
+  }
+
+  getProviderName(): string {
+    return 'Azure Communication Services';
+  }
+
+  async validateConfig(): Promise<boolean> {
+    try {
+      // Connection string format validation
+      if (!this.connectionString || !this.connectionString.includes('endpoint=')) {
+        return false;
+      }
+      return true;
+    } catch (error) {
+      console.error('Azure Communication Services configuration validation failed:', error);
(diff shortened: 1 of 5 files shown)


gorkem-bwl:
@MuhammadKhalilzadeh fyi. This has been tested and works. Leaving the review to you.

Servers/services/email/providers/AzureCommunicationServicesProvider.ts
MuhammadKhalilzadeh: Thank you so much @Br0wnHammer 
This is a great code file. Here's a quick analysis of this files strengths:

Top 3 strengths

1. Clear, focused encapsulation: The provider is a single class with a small public surface (getProviderName, validateConfig, sendEmail). It cleanly encapsulates client creation and send logic, making it easy to swap providers or test.
2. Handles long-running send operations & normalized results: Uses the SDK's poller (beginSend + pollUntilDone) and maps the outcome to a consistent EmailResult shape (success/messageId or error object). That gives callers a simple, uniform contract.
3. Attachment and error handling attention: Converts attachment content to base64 when needed and sets a contentType fallback. Errors are caught and returned with a name/message structure instead of throwing, which is useful for higher-level flows that expect result objects.

Also, I got two suggestions as well. Please check and see if it's applicable and makes sense to apply:

1. Stronger configuration validation and secret safety
   - Current validateConfig only checks that the connection string contains "endpoint="; Azure connection strings typically also include an access key (e.g., "accesskey="). Consider validating the presence of both parts, or attempt a lightweight SDK call (or client parsing) to verify the credentials. Also avoid logging the connection string and ensure it's only provided from secure sources (env/secret manager).
   - Example improvements:
     - Check for both endpoint and accesskey substrings.
     - Return an explanation when validation fails (e.g., why) or throw a specific validation error type.

2. Sender/recipient, attachment and API-compatibility robustness
   - Sender fallback to process.env.EMAIL_ID || '' can allow an empty sender; validate and fail-fast if sender is missing. Use a clear env var name like EMAIL_FROM and surface a helpful error message.
   - Recipient handling is limited to a single "to" address; consider supporting arrays, cc, bcc, and reply-to if EmailOptions allow them.
   - Attachment conversion assumes content is string or Buffer and calls toString('base64') — be explicit about accepted types (string | Buffer | Uint8Array) and normalize accordingly (Buffer.from(...)). Also confirm the shaped attachment property (name/contentType/contentInBase64) matches the SDK's expected schema.
   - Add unit tests or integration tests to ensure mapping (message shape, attachments) matches the version of @azure/communication-email in use.

PR #2973, opened 2025-12-26
Refactor: add file repository layer for consolidated data access

gorkem-bwl:
## Summary
- Adds a consolidated file repository that combines operations from `fileUpload.utils.ts` and `fileManager.utils.ts`
- Implements the repository pattern for all file-related database operations
- Adds transaction support for atomic file operations

## Changes
- `Servers/repositories/file.repository.ts`: New repository with methods for file upload, download, metadata management, and access logging
- `Servers/repositories/index.ts`: Export barrel for repository layer

## Key features
- Centralized file operations with consistent error handling
- Transaction support for atomic operations (upload + metadata + access log)
- Filename sanitization to prevent path traversal attacks
- Support for multiple file sources (assessments, compliance, reports, etc.)
- File access logging for audit trails

## Merge order
This is PR **2 of 6** in the refactoring series. Merge in this order:
1. #2972 - Controller wrapper pattern
2. **#2973** - File domain repository (this PR)
3. #2974 - Change history consolidation
4. #2975 - Replace any types
5. #2976 - Deprecate entity change history files
6. #2977 - Vendor service layer

### DIFF
--- a/Servers/repositories/file.repository.ts
+++ b/Servers/repositories/file.repository.ts
@@ -50,10 +50,9 @@ export interface UploadedFile {
 export interface FileMetadata {
   id: number;
   filename: string;
-  size?: number;
-  mimetype?: string;
-  upload_date?: Date;
-  uploaded_by: number;
+  project_id?: number;
+  uploaded_time?: string;
+  source?: string;
   uploader_name?: string;
   uploader_surname?: string;
 }


Servers/repositories/file.repository.ts
MuhammadKhalilzadeh, line 1: Analysis of Servers/repositories/file.repository.ts

Summary
- The file centralizes file-related DB operations and includes helpful validation, filename sanitization, and types.
- Overall approach is reasonable, but I found several correctness, consistency, and safety issues you should address before merging.

Major issues and suggested fixes

1) Inconsistent / incorrect handling of sequelize.query results
- Problem: The code mixes mapToModel/model with raw QueryTypes and inconsistent result handling (e.g., uploadProjectFile checks Array.isArray(result[0]) etc.). sequelize.query returns different shapes depending on options:
  - For QueryTypes.SELECT you get an array of rows.
  - For INSERT/DELETE with RETURNING in Postgres you typically get an array of returned rows (as SELECT).
  - Using mapToModel/model returns model instances, but shape is still [results, metadata] in many versions.
- Risk: wrong assumptions can cause runtime errors or return wrong values.
- Fix: pick one consistent pattern. For Postgres + RETURNING, use QueryTypes.SELECT (or no type) and destructure the first element. Example:

  const rows = await sequelize.query(query, {
    replacements,
    type: QueryTypes.SELECT,
    transaction
  });
  return rows[0] ?? null;

  Or when you want model instances, use model + mapToModel and extract result[0] (which is array of instances):
  const instances = await sequelize.query(query, { model: FileModel, mapToModel: true, transaction });
  return instances[0] ?? null;

  Apply consistently to uploadProjectFile, deleteProjectFileById, getProjectFileById, deleteFileById.

2) deleteProjectFileById return value is unreliable
- Problem: deleteProjectFileById uses mapToModel/model but doesn't parse result correctly; returns result.length > 0 which inspects outer array, not returned rows.
- Fix: use QueryTypes.SELECT with RETURNING id and then return (rows && rows.length > 0).

Example:
  const rows = await sequelize.query(query, {
    replacements: { id },
    type: QueryTypes.SELECT,
    transaction
  });
  return Array.isArray(rows) && rows.length > 0;

3) Table identifier quoting is inconsistent
- Examples:
  - Many queries use ${escapePgIdentifier(tenant)}.files (unquoted "files")
  - getFileById builds table as '"file_manager"' or '"files"' then uses ${escapePgIdentifier(tenant)}.${table}
- While those variants produce valid SQL, the inconsistency is confusing and error-prone.
- Recommendation: keep table names as constants (no variable quoting) and consistently call escapePgIdentifier only for the schema (tenant). Eg:
  const schema = escapePgIdentifier(tenant);
  const table = '"file_manager"'; // keep constant if you need quoting
  const sql = `SELECT * FROM ${schema}.${table} WHERE id = :fileId`;

4) Escaping identifiers / validateTenant
- validateTenant limits tenant to /^[A-Za-z0-9_]{1,30}$/. This is good as a safety measure to prevent SQL injection via schema names.
- Note: some legitimate tenant/schema names may include hyphens or dots; if your system allows them you’ll need a different validation or to map tenant names to safe schema names.
- Also, escapePgIdentifier currently calls validateTenant and then does replace(/"/g, '""') and wraps in quotes — fine given validateTenant, but you only ever call it for tenant. If you later allow dynamic table names you must escape them too (and validate).

5) Using raw SQL for schema-scoped queries — consider alternatives
- Building SQL with interpolated schema is necessary (Postgres doesn't parameterize identifiers), but consider:
  - Using Sequelize's schema() feature (Model.schema(schemaName)) to avoid hand-creating SQL and reduce risk.
  - Using a white-list mapping tenant -> schema name if tenants are application-controlled.

6) Replacements containing undefined
- Several queries add replacements for limit/offset even when undefined. Sequelize ignores unused replacements in many cases, but to be safe only pass replacements keys that are actually used.
- Example: build a replacements object and only add limit/offset when provided.

7) Inconsistent interfaces for uploaded files
- uploadProjectFile expects UploadedFile interface; uploadFileManagerFile expects Express.Multer.File. Consider unifying to a single UploadedFile typed to optional buffer/size to make repository layer framework-agnostic.

8) Schema/column name mismatches vs TypeScript types
- getProjectFileMetadata SELECT returns project_id, uploaded_time, source but FileMetadata interface does not include those fields (interface has size, mimetype, upload_date, uploaded_by). Keep types aligned with returned columns or map rows to the interface explicitly.

9) Error handling
- Most functions do not catch DB errors and wrap them in domain-specific exceptions. Consider catching DB errors and throwing ValidationException or a RepositoryException with context so callers can handle predictable errors.

10) Storing file content in DB implications
- This code stores file content in content column (bytea). If files can be large, consider:
  - A max file size check and rejecting oversized files earlier.
  - Storing large objects in S3 or other object storage and storing only references/paths in DB.

Minor issues / improvements
- sanitizeFilenameStr fallback uses name.replace(/[^a-zA-Z0-9-_\\.]/g, "_") — OK, but consider truncating long filenames (e.g., >255).
- Mixed timestamp usage: uploadProjectFile sets uploaded_time with new Date().toISOString() while uploadFileManagerFile uses NOW() in SQL. Pick one approach for consistency.
- getOrganizationFiles: COUNT returns count as string — you already parseInt, good.
- Typing: some functions return Promise<FileModel | FileManagerMetadata | null>, ensure callers handle both shapes.
- Tests: add unit/integration tests for:
  - validateTenant rejection cases
  - upload & retrieval flow (including buffer roundtrip)
  - delete functions returning correct boolean
  - file access logs insertion and pagination
  - edge cases: tenant invalid, file not found, large buffer

Concrete suggested fixes (high-level)
- Standardize query result handling. Example pattern for queries that return rows:

  const rows = await sequelize.query(query, {
    replacements,
    type: QueryTypes.SELECT,
    transaction
  });
  return rows as ExpectedRowType[];

- Fix deleteProjectFileById:

  export async function deleteProjectFileById(...): Promise<boolean> {
    validateTenant(tenant);
    const query = `DELETE FROM ${escapePgIdentifier(tenant)}.files WHERE id = :id RETURNING id`;
    const rows = await sequelize.query(query, {
      replacements: { id },
      type: QueryTypes.SELECT,
      transaction,
    });
    return Array.isArray(rows) && rows.length > 0;
  }

- Fix uploadProjectFile to return the actual created row:

  const rows = await sequelize.query(query, {
    replacements: {...},
    type: QueryTypes.SELECT,
    transaction
  });
  return rows[0] as unknown as FileModel;

- Consolidate file interfaces and use a single UploadedFile type for both upload functions.

- Consider switching to model methods with Model.schema(tenant) or Sequelize transactions to avoid manual identifier construction.

Risk summary (security & correctness)
- SQL injection risk mitigated for tenant due to validateTenant, but be careful if later accepting arbitrary identifiers (table names).
- Inconsistent result parsing can cause runtime failures or incorrect return values.
- Potential unhandled large payloads and missing size validation.


CC: @gorkem-bwl Please focus on the comments above

MuhammadKhalilzadeh (approved):
No problems in file upload

PR #3005, opened 2026-01-01
refactor: replace any types with proper TypeScript interfaces in repository layer

gorkem-bwl:
## Summary

- Replace all `any` types with proper TypeScript interfaces across 45 repository files
- Add `BackendResponse<T>` and `ApiResponse<T>` wrapper types for consistent API typing
- Use `error: unknown` instead of `error: any` in all catch blocks with proper type assertions
- Add domain-specific interfaces for each repository (e.g., `Project`, `Vendor`, `Task`, etc.)

### New shared types (`ApiTypes.ts`)
- `BackendResponse<T>` - Standard wrapper `{ message: string, data: T }`
- `ApiResponse<T>` - Axios response wrapper `{ data: T, status, statusText, headers? }`

### Repository layer improvements
- All GET functions returning `response.data` now return `BackendResponse<T>`
- All POST/PATCH/PUT functions returning full `response` now return `ApiResponse<BackendResponse<T>>`
- All DELETE functions use `ApiResponse<null>` when returning full response
- All error handling uses `error: unknown` with proper type assertions
- Removed all `eslint-disable @typescript-eslint/no-explicit-any` comments (except `entity.repository.ts` which is intentionally generic)

### Files modified
- 42 repository files in `Clients/src/application/repository/`
- `Clients/src/domain/types/ApiTypes.ts` (new shared types)
- `Clients/src/domain/types/VendorRisk.ts` (new types)
- `Clients/src/infrastructure/api/networkServices.ts` (ApiResponse export)

### DIFF
--- a/Clients/src/application/hooks/useProjects.ts
+++ b/Clients/src/application/hooks/useProjects.ts
@@ -16,8 +16,8 @@ export const useProjects = () => {
   return useQuery({
     queryKey: projectQueryKeys.list(),
     queryFn: async (): Promise<Project[]> => {
-      const response = await getAllProjects();
-      return response?.data || [];
+      const projects = await getAllProjects();
+      return projects || [];
     },
     staleTime: 5 * 60 * 1000, // 5 minutes
     gcTime: 10 * 60 * 1000,   // 10 minutes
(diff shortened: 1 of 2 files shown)


Clients/src/application/repository/assesment.repository.ts
MuhammadKhalilzadeh: We have `unknow` type used in different places in here as well ❌

Clients/src/application/hooks/useAITrustCentreOverview.tsx
MuhammadKhalilzadeh: We're using `unknown` in here as well.
We can either remove them, or replace with real data types

PR #3012, opened 2026-01-04
fix: add tenant schema format validation for SQL injection defense

gorkem-bwl:
## Summary

- Adds defense-in-depth validation for tenant schema names used in SQL queries
- Validates tenant format in auth middleware before setting `req.tenantId`

## Problem

Tenant schema names are interpolated into SQL queries across 58+ utility files. While tenant values come from signed JWTs and are validated against organization hash, adding explicit format validation provides an additional security layer.

## Changes

**`Servers/utils/security.utils.ts`**
- Updated `isValidTenantHash()` to match exact format (10 alphanumeric characters)
- Added `validateTenantSchema()` utility for optional query-level validation

**`Servers/middleware/auth.middleware.ts`**
- Added format check using `isValidTenantHash()` before setting `req.tenantId`
- Rejects requests with malformed tenant values with 400 status

## Impact

- All authenticated endpoints automatically benefit from this validation
- No changes required to the 58+ query files that use tenant schemas
- Minimal performance impact (single regex check per request)

### DIFF
--- a/Servers/utils/security.utils.ts
+++ b/Servers/utils/security.utils.ts
@@ -29,28 +29,6 @@ export function isValidTenantHash(tenantId: string): boolean {
 }
 
 /**
- * Validates and returns a safe tenant schema name for use in SQL queries.
- * Throws an error if the tenant hash format is invalid.
- *
- * This is a defense-in-depth measure - tenant values come from signed JWTs
- * and are validated in auth middleware, but this provides an additional
- * layer of protection against SQL injection.
- *
- * @param tenant - The tenant hash to validate
- * @throws Error if tenant format is invalid
- * @returns The validated tenant hash
- */
-export function validateTenantSchema(tenant: string): string {
-  if (!tenant || typeof tenant !== 'string') {
-    throw new Error('Tenant schema is required');
-  }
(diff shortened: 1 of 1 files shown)


Servers/utils/security.utils.ts
```diff
@@ -17,15 +17,37 @@ export function isValidSQLIdentifier(value: string): boolean {
...
+ *
+ * This is a defense-in-depth measure - tenant values come from signed JWTs
+ * and are validated in auth middleware, but this provides an additional
+ * layer of protection against SQL injection.
+ *
+ * @param tenant - The tenant hash to validate
+ * @throws Error if tenant format is invalid
+ * @returns The validated tenant hash
+ */
+export function validateTenantSchema(tenant: string): string {
```
HarshP4585: This function is not used

gorkem-bwl: Should be fixed.

Servers/middleware/auth.middleware.ts
MuhammadKhalilzadeh, line 1: ✅

MuhammadKhalilzadeh (approved):

PR #3013, opened 2026-01-04
fix: add authorization check to file download endpoint (IDOR)

gorkem-bwl:
## Summary

- Add ownership/membership verification before allowing file downloads
- Prevent unauthorized access to files by guessing/enumerating file IDs

## Changes

**`Servers/utils/fileUpload.utils.ts`**
- Added `canUserAccessFile()` function that checks if a user has legitimate access to a file

**`Servers/controllers/file.ctrl.ts`**
- Added authorization check in `getFileContentById()` before returning file content
- Returns 403 "Access denied" if user doesn't have access

## Access rules

A user can access a file if any of the following conditions are met:
- User has Admin role
- User uploaded the file (`uploaded_by = userId`)
- User owns the project the file belongs to (`projects.owner = userId`)
- User is a member of the project (`projects_members.user_id = userId`)

For files without a project (e.g., reports), only the uploader or Admin can access them.

### DIFF
--- a/Servers/controllers/file.ctrl.ts
+++ b/Servers/controllers/file.ctrl.ts
@@ -24,6 +24,11 @@ export async function getFileContentById(
 ): Promise<any> {
   const fileId = parseInt(req.params.id, 10);
 
+  // Validate fileId is a valid number
+  if (isNaN(fileId)) {
+    return res.status(400).json({ message: "Invalid file ID" });
+  }
+
   // Validate authentication - these should be set by authenticateJWT middleware
   if (!req.userId) {
     return res.status(401).json({ message: "Unauthenticated" });


Servers/controllers/file.ctrl.ts
MuhammadKhalilzadeh, line 1: @gorkem-bwl Can you please check the following analysis with Claude Code and see if it's valuable to apply or not?

------

Problems I see
- Type error risk: the function signature still uses Express's Request, but the code reads req.userId and req.role. Unless the Request type has been extended, TypeScript will error. Use a request type that includes those fields (e.g., RequestWithFile or a dedicated AuthenticatedRequest) or cast req appropriately.
- Unsafe non-null assertions: you use req.userId! and req.tenantId!. If those can be missing at runtime this will pass undefined to canUserAccessFile or cause issues. Check for presence and return 401/400 if unauthenticated/missing tenant instead of asserting non-null.
- parseInt missing radix: parseInt(req.params.id) should be parseInt(req.params.id, 10) to avoid ambiguous parsing.
- Authorization ordering / semantics:
  - You call canUserAccessFile before fetching the file (getFileById). This can be fine if canUserAccessFile implements its own existence check, but it can produce surprising behavior: a non-existent file may yield 403 (Access denied) instead of 404 (Not found). Decide desired behavior and either:
    - fetch the file first, return 404 if missing, then call canUserAccessFile to authorize; or
    - keep auth-first intentionally to avoid leaking existence, but ensure canUserAccessFile handles missing files correctly.
  - Also consider efficiency: if canUserAccessFile internally fetches the file metadata, you may be duplicating DB calls — consider reusing metadata.
- Error logging: logFailure is passed new Error("Forbidden") — that hides the underlying reason. If canUserAccessFile throws or provides a reason, pass that to the logger to aid debugging.
- Response code choice: you return 403 for denied access. If the user is unauthenticated (no userId), 401 may be more appropriate; keep consistent with project's conventions.
- Assumed canUserAccessFile signature/behavior: ensure the order and types (fileId, userId, role, tenantId) match the actual implementation. If not, this will fail at runtime/compile time.

Suggested minimal changes
- Add radix:
  const fileId = parseInt(req.params.id, 10);
- Use a typed request and validate user/tenant:
  export async function getFileContentById(req: RequestWithUser /*or RequestWithFile*/, res: Response) {
    if (!req.userId) return res.status(401).json({ message: "Unauthenticated" });
    const userId = req.userId;
    const role = req.role ?? "";
    const tenantId = req.tenantId;
    if (!tenantId) return res.status(400).json({ message: "Missing tenant" });
    ...
  }
- Decide auth order: example (check existence first):
  const file = await getFileById(fileId, tenantId);
  if (!file) return res.status(404).json({ message: "File not found" });
  const hasAccess = await canUserAccessFile(fileId, userId, role, tenantId);
  if (!hasAccess) return res.status(403).json({ message: "Access denied" });
- Improve logging:
  await logFailure({ ..., error: new Error("Access denied for user X") }) or include any error from canUserAccessFile.

Servers/utils/fileUpload.utils.ts
MuhammadKhalilzadeh, line 1: CC: @gorkem-bwl 
Please do the same for the followings as well:

---------

Problems and risks
- escapePgIdentifier is not shown in this file. If that helper is not defined/imported, this will throw a ReferenceError at runtime.
- Using template interpolation for the schema (and the table name) is risky unless escapePgIdentifier properly quotes identifiers. If tenant is user-controlled and not validated/quoted, this is an SQL-injection vulnerability (identifiers cannot be parameterized with replacements).
- The current interpolation uses `${escapePgIdentifier(tenant)}.files` — that only correctly quotes the schema part. It assumes the "files" identifier does not need quoting or that escapePgIdentifier returns a value that makes the whole qualified name valid. It's safer to quote both schema and table (e.g. `${quoteIdent(tenant)}.${quoteIdent('files')}`).
- Role comparison is case-sensitive: role === "Admin". If your app uses different capitalization (e.g. "admin"), this will fail. Normalize or use an enum/constant.
- Minor: sequelize.query(QueryTypes.SELECT) returns an array of rows; you rely on result.length > 0 which is OK, but you could add a TS generic type for clarity (e.g. sequelize.query<{ id: number }>(...)).
- LEFT JOIN to projects_members may duplicate rows if there are multiple membership entries for the same user/project — LIMIT 1 prevents a functional problem but you may prefer EXISTS which is clearer and sometimes more efficient.

gorkem-bwl:
Some of them are fixed CC @MuhammadKhalilzadeh

MuhammadKhalilzadeh (approved):

PR #3065, opened 2026-01-14
fix: preserve formatting when loading policy templates

gorkem-bwl:
## Summary
- Fix policy template formatting not being preserved when opened in the editor

## Problem
When opening a policy template, the formatting (headings, bold text, etc.) was being lost. The content would display as plain text without the intended structure.

## Root cause
DOMPurify was stripping Slate editor data attributes (`data-slate-node`, `data-slate-type`, etc.) during content sanitization because they weren't in the allowed attributes list.

## Solution
Added the required Slate data attributes to the `ALLOWED_ATTR` list:
- `data-slate-node`
- `data-slate-type`
- `data-slate-id`
- `data-slate-editor`
- `data-slate-leaf`
- `data-slate-string`
- `data-slate-bold`
- `data-block-id`


gorkem-bwl:
Closing - this fix doesn't address the root cause. The issue is that templates use `<div data-slate-type="p">` instead of `<p>` tags.

PR #3079, opened 2026-01-14
fix: stop experiments early on fatal API errors

gorkem-bwl:
## Summary
- Adds early termination for experiments when fatal API errors are detected
- Creates new `error_detection.py` utility with pattern matching for fatal vs transient errors
- Stops experiment after 2 consecutive fatal errors and marks status as "failed"

## Problem
Previously, experiments would continue running all prompts even when API errors like "credit balance is too low" or "invalid API key" occurred. This wasted time and left experiments in a misleading "completed" state with all metrics showing "N/A".

## Solution
- **Fatal errors** (won't recover): insufficient credits, invalid API key, model not found, account disabled
- **Transient errors** (might recover): timeout, rate limit, connection issues
- After 2 consecutive fatal errors of the same type, experiment stops early
- Experiment status is set to "failed" with a clear error message

## Test plan
- [ ] Run experiment with valid API key - should complete normally
- [ ] Run experiment with invalid/empty API key - should stop after 2 attempts and show "failed" status
- [ ] Run experiment with low credits - should detect and stop early
- [ ] Verify partial results are preserved when experiment stops early

### DIFF
--- a/EvalServer/src/utils/error_detection.py
+++ b/EvalServer/src/utils/error_detection.py
@@ -1,0 +1,247 @@
+"""
+Fatal error detection for experiment execution.
+
+Detects API errors that are unrecoverable (won't succeed on retry)
+and should stop the experiment early to avoid wasting time.
+"""
+
+from typing import Tuple, Optional
+
+# Fatal error patterns that won't recover - experiment should stop
+# Format: (pattern_to_match, error_code, user_friendly_message)
+FATAL_ERROR_PATTERNS = [
+    # Credit/billing issues
+    (
+        "credit balance is too low",
+        "insufficient_credits",
+        "API credit balance is too low. Please add credits to your account.",
+    ),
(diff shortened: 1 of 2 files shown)


MuhammadKhalilzadeh:
@sermengi  @EfeAcar6431 Would you please take a look at this once you get the time?

sermengi:
I tested the fatal error detection fix. Key behaviors verified: 
- Experiment stops after 2 consecutive fatal errors of the same type (invalid API key, low credits, model not found, etc.)
- Transient errors (timeouts, 503s) don't trigger early termination
- Different fatal error types don't accumulate (counter resets on type change)
- Successful responses reset the fatal error counter
- Partial results are preserved when stopping early
- Works for both single-turn and multi-turn evaluation paths

sermengi (approved):

PR #3196, opened 2026-02-02
feat: add dataset inventory for EU AI Act Article 10 compliance

gorkem-bwl:
## Summary
- Add Dataset Inventory as a new tab within Model Inventory page
- Create comprehensive database schema with multi-tenant support
- Implement full CRUD backend controllers with change history tracking
- Create frontend components following existing patterns

### DIFF
--- a/Clients/src/presentation/components/Modals/NewDataset/index.tsx
+++ b/Clients/src/presentation/components/Modals/NewDataset/index.tsx
@@ -28,61 +28,17 @@ import {
   DatasetType,
   DataClassification,
 } from "../../../../domain/enums/dataset.enum";
+import {
+  NewDatasetFormValues,
+  NewDatasetFormErrors,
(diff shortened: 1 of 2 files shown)


Clients/src/application/repository/dataset.repository.ts
MuhammadKhalilzadeh, line 1: Looks fairly good and makes sense!

Clients/src/presentation/components/Modals/NewDataset/index.tsx
MuhammadKhalilzadeh, line 1: In this file, it has worked weakly on moving extracted interfaces to another file. but we can improve it on those parts

MuhammadKhalilzadeh (changes requested):
The only thing that needs improvement in some of the files, is moving interfaces to new or related ".ts" files

PR #3217, opened 2026-02-05
fix: make organization name modal mandatory on first login

gorkem-bwl:
## Summary
- Removes the "Skip for now" option from the organization name modal that appears on first login
- Users must now enter and save an organization name before they can access the dashboard
- The modal can no longer be dismissed via ESC key, X button, or backdrop click

## Changes
- **ChangeOrganizationName modal**: Removed `handleSkip` function, set `onClose` to no-op, set `showCancelButton={false}`
- **IntegratedDashboard**: Made `handleOrgModalClose` a no-op; modal only closes after successful save via `handleOrgNameSuccess`

### DIFF
--- a/Clients/src/presentation/components/Modals/StandardModal/index.tsx
+++ b/Clients/src/presentation/components/Modals/StandardModal/index.tsx
@@ -222,34 +222,36 @@ const StandardModal: React.FC<StandardModalProps> = ({
             </Stack>
             <Stack direction="row" spacing={1} alignItems="center">
               {headerActions}
-              <Box
-              component="span"
-              role="button"
-              tabIndex={0}
-              onClick={(e) => {
-                e.stopPropagation();
-                onClose();
-              }}
-              onKeyDown={(e) => {
-                if (e.key === "Enter" || e.key === " ") {
-                  e.preventDefault();
-                  onClose();
-                }
-              }}
-              sx={{
-                cursor: "pointer",
-                color: "#98A2B3",
-                display: "flex",
-                alignItems: "center",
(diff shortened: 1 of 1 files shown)


Clients/src/presentation/pages/DashboardOverview/IntegratedDashboard.tsx
```diff
@@ -182,11 +182,11 @@ const IntegratedDashboard: React.FC = () => {
     localStorage.setItem("has_seen_org_name_modal", "true");
     localStorage.removeItem("initial_org_name");
     localStorage.removeItem("initial_org_id");
+    setShowOrgNameModal(false);
   };
 
   const handleOrgModalClose = () => {
-    setShowOrgNameModal(false);
-    handleOrgNameSuccess();
+    // No-op: users must name their organization before proceeding
```
Br0wnHammer: We can remove this function entirely then

Clients/src/presentation/components/Modals/ChangeOrganizationName/index.tsx
```diff
@@ -128,12 +124,12 @@ const ChangeOrganizationNameModal: React.FC<ChangeOrganizationNameModalProps> =
       )}
       <StandardModal
         isOpen={isOpen}
-        onClose={handleSkip}
+        onClose={() => {}}
```
Br0wnHammer, line 126: Can omit this if not a required prop.

gorkem-bwl: both should be addressed.

PR #3351, opened 2026-02-18
feat: add pending invitations table to team settings

gorkem-bwl:
## Summary
- Add a **pending invitations table** below the existing team members table on the Team settings page, giving admins visibility into outstanding invitations
- Persist invitation records in a new `invitations` DB table when invite emails are sent
- Support **resend** (fresh token + re-sent email) and **revoke** (delete record) actions
- Auto-mark invitations as accepted when a user registers via the invite link

## Test plan
- [ ] Send an invitation and verify a row appears in the pending invitations section
- [ ] Resend an invitation and verify the "Sent" column updates and a new email is sent
- [ ] Revoke an invitation and verify it disappears from the table
- [ ] Register via an invite link and verify the invitation is removed from pending
- [ ] Verify expired invitations show an "Expired" chip
- [ ] Verify non-admin users cannot access invitation API endpoints (403)

### DIFF
--- a/Clients/src/application/utils/dateFormatter.ts
+++ b/Clients/src/application/utils/dateFormatter.ts
@@ -9,7 +9,19 @@ import { formatDistanceToNow, isValid } from "date-fns";
 export const formatRelativeDate = (dateString: string): string => {
   if (!dateString) return "Unknown";
 
-  const date = new Date(dateString);
+  // PostgreSQL returns timestamps without timezone info (e.g., "2026-02-19 21:30:00")
+  // We need to treat them as UTC by replacing space with 'T' and appending 'Z'
+  let normalizedDateString = dateString.trim();
+
+  // Replace space with 'T' for ISO format
+  normalizedDateString = normalizedDateString.replace(' ', 'T');
+
+  // Append 'Z' if no timezone indicator present
+  if (!normalizedDateString.endsWith('Z') && !normalizedDateString.match(/[+-]\d{2}:?\d{2}$/)) {
+    normalizedDateString += 'Z';
+  }
+
+  const date = new Date(normalizedDateString);
   if (!isValid(date)) return "Unknown";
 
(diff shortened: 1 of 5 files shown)


Servers/database/migrations/20260218200000-create-invitations-table.js
```diff
@@ -0,0 +1,38 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.sequelize.query(`
+      CREATE TABLE IF NOT EXISTS invitations (
```
HarshP4585: This is not tenant-based, we need to create `invitations` tenant-based table

Servers/controllers/invitation.ctrl.ts
```diff
@@ -0,0 +1,99 @@
...
+    console.error("Error fetching invitations:", error);
+    return res.status(500).json({ error: "Failed to fetch invitations" });
+  }
+};
+
+/**
+ * DELETE /api/invitations/:id
+ * Revoke a pending invitation.
+ */
+export const revokeInvitation = async (
```
HarshP4585, line 32: Even if we revoked invitation and the user registers, I think he will be able to create an account

HarshP4585 (changes requested):

Servers/controllers/invitation.ctrl.ts
gorkem-bwl: Yes, that is more about joiing to that organization with a specific organization ID. If we revoke, I am ok he registers again with a new organization ID. What do you think?

Servers/database/migrations/20260218200000-create-invitations-table.js
gorkem-bwl: Yes, making necessary changes now.

PR #3381, opened 2026-02-20
fix: add missing CSS and font files that break client build

gorkem-bwl:
## Summary
- Add `fonts.css` and self-hosted Geist/Red Hat Display woff2 fonts that `main.tsx` imports but were never committed
- Add `icon-shake.css` animation used by `DashboardActionButtons`, `SidebarShell`, and `NotificationBell` but never committed

## Test plan
- [ ] `npm run dev` in Clients — no Vite import resolution errors
- [ ] Font rendering uses self-hosted Geist and Red Hat Display
- [ ] Icon shake animation works on notification bell hover and sidebar icons


Br0wnHammer (commented):
LGTM!

PR #3399, opened 2026-02-23
Optimize dashboard loading performance and add progress dialog

gorkem-bwl:
## Summary
- Remove 3 duplicate/unused API calls from dashboard cold load (17-22 calls → 13-18)
- Merge duplicate `/projects` and `/modelInventory` fetches into single calls that serve multiple consumers
- Parallelize ISO framework clause/annex sub-calls instead of running them sequentially
- Replace generic loading spinner with step-based progress dialog on cold loads

## Test plan
- [ ] Clear localStorage (`dashboard_metrics_cache` key) and reload dashboard — verify step progress dialog appears with advancing steps
- [ ] Refresh within 30s — verify instant load with no progress dialog
- [ ] Open Network tab — confirm no duplicate `/projects` or `/modelInventory` calls, no `/users` call
- [ ] Verify all dashboard cards render correctly in both executive and operations views
- [ ] Navigate to AI Detection and trigger a scan recalculation — verify the timer-based `StepProgressDialog` still works (backward compatibility)

### DIFF
--- a/Clients/src/application/hooks/useDashboardMetrics.ts
+++ b/Clients/src/application/hooks/useDashboardMetrics.ts
@@ -3,7 +3,7 @@ import { getAllEntities, getEntityById } from "../repository/entity.repository";
 import type { ProgressStep } from "../../presentation/components/StepProgressDialog";
 
 // Cache configuration
-const CACHE_KEY = "dashboard_metrics_cache";
+export const CACHE_KEY = "dashboard_metrics_cache";
 const CACHE_TTL_MS = 30 * 1000; // 30 seconds - data is considered fresh
 const STALE_TTL_MS = 5 * 60 * 1000; // 5 minutes - data can still be shown while revalidating
 
@@ -68,6 +68,14 @@ const setCachedValue = <T>(key: keyof MetricsCache, data: T): void => {
   const cache = getCache();
   cache[key] = { data, timestamp: Date.now() };
   setCache(cache);
+};
+
+/**
+ * Check if any cached dashboard metrics exist in localStorage.
+ * Used by IntegratedDashboard to decide whether to show the progress dialog.
+ */
+export const hasDashboardCache = (): boolean => {
+  return Object.keys(getCache()).length > 0;
 };
 
 // Types for the additional dashboard metrics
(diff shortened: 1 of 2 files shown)


Clients/src/application/hooks/useDashboardMetrics.ts
```diff
@@ -1,5 +1,6 @@
-import { useEffect, useState, useCallback } from "react";
+import { useEffect, useState, useCallback, useRef } from "react";
 import { getAllEntities, getEntityById } from "../repository/entity.repository";
+import type { ProgressStep } from "../../presentation/components/StepProgressDialog";
 
 // Cache configuration
 const CACHE_KEY = "dashboard_metrics_cache";
```
Br0wnHammer: We can export `CACHE_KEY` as it is being used in two places: useDashboardMetrics and IntegratedDashboard

Clients/src/presentation/pages/DashboardOverview/IntegratedDashboard.tsx
```diff
@@ -62,43 +62,24 @@ type DashboardView = "executive" | "operations";
...
-      // Clear timer and hide loading indicator
-      if (loadingTimerRef.current) {
-        clearTimeout(loadingTimerRef.current);
-        loadingTimerRef.current = null;
-      }
-      setShowLoadingIndicator(false);
-      // Mark content as ready for fade-in
-      setContentReady(true);
+  // Check if there is any cached dashboard metrics data
+  const hasAnyCache = useMemo(() => {
```
Br0wnHammer: `Suggestion`: Since we already have a `getCache()` helper in useDashboardMetrics that reads from localStorage using the same `CACHE_KEY`, we could export a small utility (or a `hasAnyCache()` helper) from that module instead of re-implementing the cache check here. That would keep the cache key and parsing logic single-sourced and avoid the two implementations drifting over time.

Clients/src/application/hooks/useDashboardMetrics.ts
gorkem-bwl: Done — exported CACHE_KEY and added a hasDashboardCache() helper in useDashboardMetrics.ts. IntegratedDashboard now imports both instead of duplicating. See 15f2bca.

Clients/src/presentation/pages/DashboardOverview/IntegratedDashboard.tsx
gorkem-bwl: Good call — added hasDashboardCache() in useDashboardMetrics.ts that wraps the existing getCache() logic. IntegratedDashboard now uses it instead of the inline implementation. See 15f2bca.

PR #3514, opened 2026-03-09
feat(policies): add one-way DOCX import to policy editor

gorkem-bwl:
## Summary
- Adds a DOCX import feature to the policy editor using mammoth.js for conversion
- Consolidates two separate export buttons (PDF, Word) into a single Export dropdown
- Adds an Import button that accepts `.docx` files and converts them to TipTap-compatible HTML
- Shows a confirmation dialog when importing over existing editor content

## Details
- **Backend**: Stateless `POST /policies/import/docx` endpoint — receives DOCX via multer, converts with mammoth.js, returns sanitized HTML. No database changes.
- **Frontend**: Export dropdown (Popover), Import button with hidden file input, `ConfirmationModal` for replace warning, DOMPurify sanitization before setting editor content
- Validates both MIME type and `.docx` file extension server-side


Servers/controllers/policy.ctrl.ts
Br0wnHammer: Please remove these console logs.

Clients/src/presentation/pages/PolicyDashboard/PolicyEditorPage.tsx
```diff
@@ -1481,84 +1530,147 @@ export default function PolicyEditorPage() {
...
-                        "&:hover": {
-                          backgroundColor: "#F9FAFB",
-                          borderColor: "#98A2B3",
-                        },
-                        "&:disabled": {
-                          backgroundColor: "#F9FAFB",
-                          borderColor: "#E4E7EC",
-                          color: "#98A2B3",
-                        },
+                <CustomizableButton
```
Br0wnHammer, line 1543: Consider keeping the modal open with a loading state until import completes, using the isLoading prop that ConfirmationModal already supports.

Servers/controllers/policy.ctrl.ts
```diff
@@ -272,6 +273,52 @@ export class PolicyController {
...
+      html = html.replace(/<(\/?)h[456](\s|>)/gi, "<$1h3$2");
+      html = html.replace(/\s+class="[^"]*"/g, "");
+
+      const warnings = result.messages
+        .filter((m) => m.type === "warning")
+        .map((m) => m.message);
+
+      return res.status(200).json(STATUS_CODE[200]({ html, warnings }));
+    } catch (error) {
+      console.error("Error importing DOCX:", error);
```
Br0wnHammer: Controllers should use the structured logging helpers.

Clients/src/presentation/pages/PolicyDashboard/PolicyEditorPage.tsx
```diff
@@ -88,6 +88,7 @@ import {
   ChevronDown as ChevronDownIcon,
   ChevronUp as ChevronUpIcon,
   Search,
+  Upload,
 } from "lucide-react";
```
Br0wnHammer, line 93: Add a client-side size check in handleDocxFileSelect before calling processDocxImport. The backend multer config limits to 10 MB, but the frontend has no pre-upload size check.

Servers/controllers/policy.ctrl.ts
```diff
@@ -272,6 +273,52 @@ export class PolicyController {
     }
   }
 
+  // Import DOCX and convert to HTML
+  static async importDocx(req: Request, res: Response) {
```
Br0wnHammer, line 285: Add server-side sanitization using a library like sanitize-html or isomorphic-dompurify before returning the response. The backend at PolicyController.importDocx returns raw mammoth output directly to the client. While the frontend does sanitize with DOMPurify, defense-in-depth demands server-side sanitization too. A malicious DOCX could contain embedded scripts or dangerous HTML that mammoth faithfully converts.

Br0wnHammer (changes requested):
The feature is well-scoped and the implementation is solid overall. However, few issues ranging from security concerns to UX bugs worth addressing before merge.

MuhammadKhalilzadeh:
What's the latest status on this @gorkem-bwl ?

MuhammadKhalilzadeh (approved):
It's very nice @gorkem-bwl 
Works without error

PR #3592, opened 2026-03-24
fix(dashboard): redirect to overview after deleting demo data

gorkem-bwl:
## Summary
- After deleting demo data, the page now redirects to `/overview` instead of reloading the current URL
- Fixes 404 errors when the user was on a page referencing a deleted entity (e.g. `/project-view?projectId=5`)

## Test plan
- [ ] Create demo data
- [ ] Navigate to the demo project view page
- [ ] Delete demo data
- [ ] Verify the page redirects to `/overview` instead of showing a 404

### DIFF
--- a/Clients/src/presentation/containers/Dashboard/index.tsx
+++ b/Clients/src/presentation/containers/Dashboard/index.tsx
@@ -234,8 +234,8 @@ const Dashboard: FC<DashboardProps> = ({ reloadTrigger }) => {
 
         setShowToastNotification(false);
 
-        // Reload the page to refresh all dashboard metrics
-        window.location.reload();
+        // Navigate to overview to avoid 404s on deleted entity pages
+        window.location.href = "/overview";
       } else {
         setShowToastNotification(false);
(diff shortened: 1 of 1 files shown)


MuhammadKhalilzadeh (changes requested):
Great, it works. Thanks @gorkem-bwl 
The only problem is that once demo data is deleted, **Delete demo data** button is still there and does not turn into **Create demo data**. (On the sidebar)

gorkem-bwl:
The `fix/demo-data-redirect` branch replaces `window.location.reload()` with `window.location.href = "/overview"`, which does a full page navigation. On reload, the Dashboard container re-fetches `projects_list` via `useDashboard()` and the `hasDemoData` check at line 61-72 will correctly return `false` since the demo project no longer exists — so the sidebar should switch back to "Create demo data."

If you're still seeing the issue on this branch, it may be a timing issue where the delete hasn't fully committed before the new page loads. Let me know if it reproduces after merging.

PR #3658, opened 2026-04-02
fix(ai-gateway): guardrail catalog enable button not reflecting enabled state

gorkem-bwl:
## Summary
- Fixed guardrail catalog "Enable" button never updating to "Enabled" after activation
- Root cause: `loadRules()` read `res.data.data` but the API returns `{ rules: [...] }`, so rules were always an empty array and the catalog could never match existing rules
- Also cleaned up 13 duplicate PII rules in the database that were created by repeated Enable clicks while the bug was present

## Test plan
- [ ] Open AI Gateway > Guardrails > PII detection
- [ ] Click "Add from catalog"
- [ ] Verify already-enabled rules show "Enabled" text instead of "Enable" button
- [ ] Click "Enable" on a new rule and verify it switches to "Enabled" without page reload
- [ ] Verify the PII detection list on the main page shows all configured rules


Br0wnHammer (approved):
LGTM! Tested.

Issue #3828, opened 2026-05-01
EU AI Act: Owner/Reviewer/Approver IDs are not validated against project members


gorkem-bwl:
**Severity:** Low (data integrity)

**Symptom**

The EU AI Act control save endpoint accepts any user ID for the Owner / Reviewer / Approver fields without checking:

- whether the ID corresponds to an existing user in the org,
- whether the user is a member of the project the control belongs to,
- whether the user has the role (Reviewer / Approver) appropriate for the assignment.

The backend stores whatever ID the client sends (subject only to the FK constraint to `users.id`).

**Why it matters**

A frontend bug, a stale list, or a malicious client could assign controls to users who shouldn't have access — silently corrupting accountability for compliance items.

**Suggested fix**

In `saveControls`, validate each non-null Owner/Reviewer/Approver against `projects_members` and against the appropriate `Role`. Return a 400 with a descriptive message if validation fails.

**Related**

Discovered during the fix/eu-control-owner-rollup review. Pre-existing — the same gap existed for subcontrol-level assignments before the rollup. Now more visible because assignments live at the control level.

Issue #3827, opened 2026-05-01
EU AI Act: Owner/Reviewer/Approver/Risk review selects have no clear option


gorkem-bwl:
**Severity:** Low (UX gap)

**Symptom**

In the EU AI Act control drawer (Use case → Frameworks → EU AI Act → Requirements → click a control), the Owner / Reviewer / Approver / Risk review dropdowns only list the actual options. There is no "(none)" or clear button.

Once an Owner is set, the user cannot unset it through the UI — only switch to a different person. The same applies to Reviewer, Approver, and Risk review.

**Backend supports clearing**

`updateControlEUByIdQuery` correctly persists `null` for these fields when called with `null`, so the data layer is ready. The gap is purely in the UI.

**Suggested fix**

Add a clear-option (e.g. an "Unassign" entry at the top of the user-list dropdowns, or a small × button next to the control). Same component is reused across other frameworks — fix once, benefit everywhere.

**Related**

Discovered while testing fix/eu-control-owner-rollup. Not a regression — pre-existing behavior, but more visible now that ownership lives at the control level.

Issue #3826, opened 2026-05-01
EU AI Act: due_date drifts one day per save/reload cycle (timezone bug)


gorkem-bwl:
**Severity:** Medium (data corruption on every save)

**Symptom**

Setting a due date on an EU AI Act control and saving, then reopening the drawer, shows the date one day earlier. Saving again shifts it another day. Reproduced consistently in dev:
- Entered: 12/31/2026
- After 1st save/reload: 12/30/2026
- After 2nd save/reload: 12/29/2026

**Root cause (suspected)**

Postgres column `controls_eu.due_date` is `TIMESTAMP WITH TIME ZONE`. The frontend formats the picker value as `YYYY-MM-DD` (string with no time component). When Postgres stores a bare date as `TIMESTAMP WITH TIME ZONE`, it interprets midnight in the *session* timezone. On readback, the API returns a full ISO datetime, and JS `new Date('YYYY-MM-DD')` treats bare dates as **UTC midnight**, dropping a day in any negative-offset timezone.

The bug is not specific to the EU AI Act framework — it likely affects all date pickers in the app that round-trip through this column type.

**Reproduce**

1. Open a control on an EU AI Act use case.
2. Set a due date and save.
3. Close drawer, reopen.
4. Date is shifted one day earlier.
5. Save again — drifts another day.

**Suggested fix**

- Change column type to `DATE` (no timezone) for date-only fields. OR
- Send/receive as a discrete `{year, month, day}` payload, not a string. OR
- Normalize on the frontend: when reading a date string, parse it as local-midnight rather than UTC.

**Related**

Discovered while reviewing fix/eu-control-owner-rollup branch. Not introduced by that PR — same drift exists for any date field stored in `TIMESTAMP WITH TIME ZONE` columns.

PR #3829, opened 2026-05-01
EU AI Act: move owner/reviewer/approver/due date/risk review to the control level

gorkem-bwl:
## Summary

EU AI Act controls displayed "Not set" for Owner and "Waiting" in the Use Case report no matter what the user filled in. The save path stopped writing to `controls_eu` in a previous refactor while the read path (controls list, report) continued to read from those same columns. Fields lived only on subcontrols and the Control row was always blank.

This PR consolidates ownership at the Control level and derives the Control's status from its subcontrols.

## Changes

- **Migration** rolls up the most-common non-null subcontrol value for `owner`, `approver`, `reviewer`, `due_date`, `risk_review` into the parent `controls_eu` row (only when the parent column is null), then nulls out those columns on subcontrols. Snapshot tables preserve the pre-migration state for rollback. Status is also derived during the migration.
- **`saveControls`** now persists the control-level fields, derives status from subcontrol statuses on every save, and fires assignment notifications at the Control level instead of per-subcontrol.
- **Reads** (`getCompliancesEUByIdQuery`, `getControlStructByControlCategoryIdForAProjectQuery`) keep selecting from `controls_eu`, which is now actually populated.
- **`updateControlEUByIdQuery`** truthy-check loosened so callers can clear a value (`null`).
- **Frontend** moves Owner / Reviewer / Approver / Due date / Risk review from per-subcontrol to control-level inputs in the drawer. Subcontrols keep Status, Implementation description, Evidence, Cross mappings, Notes.
- **Status derivation** rule: all subcontrols Done → Done; any in progress (or partial Done) → In progress; otherwise Waiting.
- **Empty-string normalization** in the controller turns blank form fields into null so Postgres doesn't reject `''::timestamp` or `''::enum`.
- **Status derivation source** is the full set of subcontrol rows in DB, not just the ones in the request payload (avoids partial-submit corruption).
- **`ControlEU` type** widened to allow `null` for owner/reviewer/approver/due_date/risk_review/status/implementation_details so the type matches the persistence layer's clear-by-null contract.

## Verified locally

- Round-trip on the dev DB: 184 → 39 expected, 1 of 2 subcontrols Done → "In progress", both Done → "Done"
- Owner column populates after save
- Notifications fire on assignment
- `npm run build` / format-check pass on both Servers and Clients
- 270 existing tests + 4 new `deriveControlStatus` tests pass

## Known follow-ups (filed as separate issues)

- #3826 — `due_date` drifts one day per save (timezone bug, pre-existing)
- #3827 — Selects have no clear option (pre-existing UX gap)
- #3828 — Owner/Reviewer/Approver IDs not validated against project members (pre-existing)


MuhammadKhalilzadeh (approved):
Things are working and look good
Thanks @gorkem-bwl

PR #3834, opened 2026-05-01
Allow clearing Owner/Reviewer/Approver in ISO 42001, ISO 27001, and NIST drawers

gorkem-bwl:
## Summary

Once a value was set on the Owner / Reviewer / Approver dropdowns in the ISO 42001, ISO 27001, and NIST AI RMF framework drawers, there was no way to unset it through the UI. This mirrored the EU AI Act control bug fixed earlier and shared the same root cause: an empty string was silently ignored by both the frontend submit guards and the backend update util.

## Changes

**Frontend (5 drawers)**

- Prepend a `(none)` entry to the Owner/Reviewer/Approver Selects so the user has a way to clear an assignment.
- Drop truthy guards in the FormData submit so an empty string for the cleared field actually reaches the backend.

**Backend (3 utils)**

- Update queries now treat empty string and null as "set this column to NULL", instead of silently skipping the field.
- Same anti-pattern existed verbatim in `iso42001.utils.ts` (`updateAnnexCategoryQuery`, `updateSubClauseQuery`), `iso27001.utils.ts` (`updateAnnexControlQuery`, `updateSubClauseQuery`), and `nist_ai_rmf.subcategory.utils.ts` (`updateNISTAIRMFSubcategoryByIdQuery`).

Status fields are intentionally left alone — every status enum value is meaningful, and `Not started` already serves as the natural reset.

## Verified locally

Verified all 5 backend endpoints round-trip correctly via direct API tests (set Owner = userId → owner = userId; clear Owner = "" → owner = null; transaction commits; DB rows match):

| Endpoint | Set | Clear |
|---|---|---|
| `iso-42001/saveClauses` | ✅ | ✅ |
| `iso-42001/saveAnnexes` | ✅ | ✅ |
| `iso-27001/saveClauses` | ✅ | ✅ |
| `iso-27001/saveAnnexes` | ✅ | ✅ |
| `nist-ai-rmf/subcategories` | ✅ | ✅ |

`npm run build` and format-check pass for both Servers and Clients. 109 jest tests pass.

## Known follow-ups

- The EU AI Act ControlPane fix and these 5 drawers all duplicate a `memberOptions` useMemo block. Worth extracting a shared `useMemberOptions(projectMembers)` hook in a follow-up.
- Pre-existing SQL bug in the ISO 27001 saveClauses notifier path: it references `clauses_struct_iso27001.arrangement` which doesn't exist. Crashes the backend on assignment notifications. Not introduced here; needs a separate fix.

PR #3837, opened 2026-05-02
Fix ISO 27001 notifier query referencing non-existent arrangement column

gorkem-bwl:
## Summary

Assigning Owner / Reviewer / Approver to an ISO 27001 subclause crashed the backend with:

```
ERROR: column "arrangement" does not exist
```

The assignment notifier in `iso27001.ctrl.ts` issues a SELECT to build the entity name for the in-app notification. The query referenced `clauses_struct_iso27001.arrangement`, but that column doesn't exist — the human-readable clause identifier ("4", "5", etc.) lives in the `clause_id` column.

## Fix

Replaced `c.arrangement` with `c.clause_id` in the subclause notifier query, preserving the `clause_arrangement` alias so the result shape and downstream formatters at lines 90 and 94 are unchanged.

## Verified locally

A PATCH to `/iso-27001/saveClauses/2` with a new owner now returns 200 and the backend stays alive — the exact scenario that previously crashed the process. TypeScript and prettier checks pass.

## Notes

The annex notifier query (line 110) uses `a.arrangement` against `annex_struct_iso27001`, which **does** have an `arrangement` column. That path was always working and is left untouched.

Discovered while reviewing the ISO/NIST clear-option PR (#3834). Pre-existing on develop.

### DIFF
--- a/Servers/controllers/iso27001.ctrl.ts
+++ b/Servers/controllers/iso27001.ctrl.ts
@@ -66,15 +66,14 @@ async function notifyIso27001Assignment(
     let description: string | undefined;
 
     if (entityType === "ISO 27001 Subclause") {
-      // Query for parent clause info, subclause order_no for full identifier (e.g., "4.1 Understanding the organization"), and subclause description
       const result = await sequelize.query<{
         clause_id: number;
-        clause_arrangement: number;
+        clause_arrangement: string;
         clause_title: string;
         subclause_order_no: number;
         requirement_summary: string;
       }>(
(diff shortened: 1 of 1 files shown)


gorkem-bwl:
@MuhammadKhalilzadeh can you check this one? That will go into 2.4 early next week.

MuhammadKhalilzadeh:
> @MuhammadKhalilzadeh can you check this one? That will go into 2.4 early next week.

Sure @gorkem-bwl 
I'll check it today 👍🏻

MuhammadKhalilzadeh:
I took care of it in a previous PR ✅

gorkem-bwl:
> I took care of it in a previous PR ✅

Thank you.

PR #3932, opened 2026-05-20
perf(loader): inline splash and prune heavy chunks from modulepreload

gorkem-bwl:
## Summary
- Inline a tiny splash spinner inside `#root` in `index.html` so the visible white-screen is replaced by a centered spinner that paints immediately, before any JS executes. The splash uses VerifyWise primary color (#13715b) and respects dark mode. React replaces it on mount.
- Add `build.modulePreload.resolveDependencies` in `vite.config.ts` to drop `vendor-charts`, `vendor-editor`, `xlsx`, and `ExportMenu` from the modulepreload list. These chunks are only needed deep inside specific pages (policy editor, exports) and were competing for bandwidth with the critical-path bundles on first paint.

## Why
On `test.verifywise.ai` the browser downloads ~1.7 MB of JS over the wire before React can mount. The most visible symptom is a white screen for a few seconds. Two issues compound it:

1. The HTML body is `<div id="root"></div>` — nothing paints until React mounts.
2. The modulepreload list eagerly fetches recharts (~162 KB gzipped) and tiptap (~139 KB gzipped) on every page load, even though almost no first-paint screen uses them.

This PR addresses both with minimal risk.

## Expected impact
- Eliminates the perceived white screen (LCP improves even if total TTI is unchanged).
- Frees ~301 KB of competing gzipped bandwidth at first paint, helping the critical chunks land faster.

### DIFF
--- a/Clients/index.html
+++ b/Clients/index.html
@@ -84,7 +84,7 @@
       }
     </style>
     <div id="root">
-      <div id="vw-splash" aria-hidden="true">
+      <div id="vw-splash" role="status" aria-live="polite" aria-label="Loading">
         <div class="vw-spinner"></div>
       </div>
     </div>
--- a/Clients/index.html
+++ b/Clients/index.html
@@ -49,7 +49,45 @@
         document.documentElement.classList.add("dark-mode");
       }
     </script>
-    <div id="root"></div>
+    <style>
+      #vw-splash {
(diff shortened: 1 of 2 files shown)


MuhammadKhalilzadeh (approved):
Thank you @gorkem-bwl

PR #4181, opened 2026-06-24
perf(dashboard): parallelize first-login metrics fetch and dedup /projects

gorkem-bwl:
## Summary

Speeds up the first-login dashboard, which loaded slowly because its metric fetches ran as a sequential waterfall with nothing on screen while waiting.

## Changes

- **Parallelize the metrics fetch.** The five metric groups (risks, vendors, models, frameworks, scores) are independent but were awaited one after another, so first-login load waited on the sum of each stage's slowest call. They now run as a single parallel batch. The framework chain (`/projects` → `/frameworks` → per-framework progress) is an internal waterfall and remains the long pole; everything else resolves underneath it.
- **Deduplicate `/projects`.** On first login the dashboard shell and the metrics hook both request `/projects` at the same moment. A new in-flight GET helper shares the promise across concurrent callers, collapsing them into one round trip. The entry clears on settle, so it only dedupes genuinely concurrent calls and never serves stale data.
- **Progress dialog stays honest.** `progressStep` tracks the leading run of completed stages, so the dialog label and percentage remain monotonic and match the work actually in flight even when groups settle out of order.
- **Abortable calls are never deduped.** Calls passing an `AbortSignal` always get their own request, so aborting one caller cannot cancel another and a single rejection does not fan out.
- **Fresh post-mutation refresh.** `getAllProjects` gains a `forceFresh` opt-out used by the dashboard's post-mutation refresh, so it cannot resolve to an older `/projects` request still in flight that predates the change.
- **Single cache flush.** Metric cache writes buffer in memory during the batch and persist once after it completes, instead of re-serializing the whole localStorage blob on each of ~18 writes.

## Benefits

- First-login dashboard drops from seven serial round trips to roughly two, a latency-bound load reduction of about 70%. No backend or API changes; payload sizes are unchanged.
- New unit tests cover the dedup helper: concurrent collapse, re-fetch after settle, params keying, signal bypass, shared rejection, and entry cleanup.

### DIFF
--- a/Clients/src/application/hooks/useDashboardMetrics.ts
+++ b/Clients/src/application/hooks/useDashboardMetrics.ts
@@ -1172,34 +1172,42 @@ export const useDashboardMetrics = () => {
       setProgressStep(0);
 
       try {
-        // Group 1: Core - risks, evidence files
-        await Promise.allSettled([fetchRiskMetrics(), fetchEvidenceMetrics()]);
-        setProgressStep(1);
+        // These metric groups are independent of one another, so they run as a
+        // single parallel batch instead of five sequential stages. On first
+        // login (empty cache) this is the difference between waiting on the sum
+        // of each stage's slowest call and waiting on the slowest call overall.
+        // Group 4 (projects -> frameworks -> per-framework progress) is an
+        // internal waterfall and stays the long pole; everything else resolves
+        // underneath it. progressStep still advances per group for the dialog,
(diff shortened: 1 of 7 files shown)


MuhammadKhalilzadeh (approved):

MuhammadKhalilzadeh (approved):
Works error-less @gorkem-bwl 👍🏻
Good job 🤝

gorkem-bwl:
Thanks!

PR #4185, opened 2026-06-25
fix(auth): reset by stored email, trim lookups, preserve anti-enumeration

gorkem-bwl:
## Summary

Fixes a case where a user could neither log in nor receive a password-reset email, while the reset screen still reported success.

## Root cause

`resetPassword` resolved the user row case-insensitively but then passed the **raw request email** into `resetPasswordQuery`, whose `UPDATE` matches `email` exactly. When the stored value differed in case or whitespace from what the user typed, the `UPDATE` matched zero rows and the password silently stayed unchanged. The reset request still returned success, so the user saw a confirmation but could never actually log in.

## Changes

- **Reset by the stored email.** `resetPassword` now passes the canonical `_user.email` into the update instead of the raw request email, so reset works for every existing row regardless of how it was created.
- **Lookups tolerate stored whitespace.** `getUserByEmailQuery` now compares `LOWER(TRIM(...))` on both sides, so existing rows with stray whitespace are found by login, the reset send path, and the duplicate-user guard, with no data migration. This also closes a path where a whitespace-padded email skipped the pre-insert duplicate check.
- **Email normalized on write.** `createNewUser` and `updateCurrentUser` trim the email before validating and storing, so future rows are canonical.
- **Forgot-password preserves anti-enumeration.** The handler awaits the request so the button stays disabled until it settles, but navigates to the confirmation screen regardless of outcome. Surfacing send failures would re-expose a found-vs-not-found oracle (the backend returns 200 for unknown accounts but 500 when an account exists and SMTP fails). It also keeps the entered email and guards against double-submit.

## Tests

- `user.ctrl`: reset updates by the stored email, not the raw request email.
- `user.model.spec`: surrounding whitespace is trimmed from the email on create.
- `ForgotPassword`: trims the email before sending, and navigates to the confirmation screen even when the request fails.


MuhammadKhalilzadeh (approved):
Looks mostly good to me @gorkem-bwl 👍🏻
What do you think @HarshP4585 ?

PR #4304, opened 2026-07-21
RLS, Auth, HTTPS & DAST/CI Scanning

MuhammadKhalilzadeh:
## Please ensure all items are checked off before requesting a review:

- [x] I deployed the code locally.
- [x] I have performed a self-review of my code.
- [x] I have included the issue # in the PR.
- [x] I have labelled the PR correctly.
- [x] The issue I am working on is assigned to me.
- [x] I have avoided using hardcoded values to ensure scalability and maintain consistency across the application.
- [x] I have ensured that font sizes, color choices, and other UI elements are referenced from the theme.
- [x] My pull request is focused and addresses a single, specific feature.
- [x] If there are UI changes, I have attached a screenshot or video to this PR.
- [x] If I added or modified an API endpoint, the change is reflected in the generated OpenAPI spec (`npm run generate:swagger`).
- [x] If the endpoint requires authentication, it uses `authenticateJWT` and the generated spec declares `bearerAuth` security.
- [x] I ran `npm run check:api-drift` and committed the regenerated `swagger.yaml` and `endpoints.ts`.
- [x] If this PR adds or modifies an organization-scoped table, the [tenant isolation registry](Servers/tests/integration/tenant-isolation/tenantIsolation.registry.ts) and [test matrix](Servers/tests/integration/tenant-isolation) are updated. See the [tenant isolation runbook](docs/technical/security/tenant-isolation.md) for details.

----------------

| Item | Status |
|---|---|
| C1 Privilege escalation + arbitrary user deletion | ✅ Fixed (`authorize(["Admin","SuperAdmin"])` on DELETE; self/admin + role-change guards in `updateUserById`) |
| C2 Slack webhook IDORs | ✅ Fixed (JWT-scoped listing; owner checks on get/update/send) |
| C3 EvalServer unauthenticated | ✅ Fixed (fail-closed `x-internal-key` middleware + startup guard + proxy + compose/ansible/install wiring of `EVAL_SERVER_INTERNAL_KEY`) |
| C4 Plain-HTTP ports public in prod compose | ✅ Fixed (backend + frontend bound to `127.0.0.1`) |
| C5 `NODE_ENV` not guaranteed in prod image | ✅ Fixed (`ENV NODE_ENV=production` in `Servers/Dockerfile`) |
| H1 HSTS + security headers at edges | ✅ Done (ansible nginx template incl. `X-Forwarded-Proto`; `Clients/nginx.conf`; k8s ConfigMap, deprecated `X-XSS-Protection` removed) |
[...]

PR #4313, opened 2026-07-22
docs(api): regenerate OpenAPI spec to fix drift

gorkem-bwl:
Targets the `mo-374-jul-20-advanced-security-scanning` branch (PR #4304), resolving the `check:api-drift` failure in backend-checks.

## Problem

The security branch added `POST /api/organizations/setup` and applied `authorize()` role gating to a number of write endpoints, but the generated `swagger.yaml` and `endpoints.ts` were not regenerated. `check:api-drift` therefore failed: 703 Express endpoints vs 702 documented operations.

## Changes

Regenerated both artifacts from the route layer (`npm run generate:swagger && npm run generate:endpoints`):

- Adds the `/organizations/setup` operation.
- Records the new role requirements (`Requires role: Admin`, `Requires role: Admin or SuperAdmin`, etc.) that the `authorize()` middleware now enforces on user-delete (the C1 privilege-escalation fix), SSO config, organization, and file-upload routes — so the spec accurately documents the PR's auth hardening.

## Result

`check:api-drift` now reports **703 endpoints == 703 operations, no drift**. Diff is limited to the two generated files, no unrelated churn and no removed operations.

### DIFF
--- a/Servers/swagger.yaml
+++ b/Servers/swagger.yaml
@@ -7890,7 +7890,7 @@ paths:
       operationId: updateUserById
     delete:
       summary: 'Delete user by ID'
-      description: "Deletes a user and nullifies all their foreign key references across\nprojects, vendors, risks, vendor risks, files, automations, and invitations.\nAlso removes the user from projects_members. Demo users and super-admins\ncannot be deleted.\n"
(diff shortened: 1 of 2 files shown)


gorkem-bwl:
FYI @MuhammadKhalilzadeh

PR #4548, opened 2026-08-24
fix SSO config endpoints crashing with config.toJSON is not a function

### LINKED ISSUE #4219: SSO config endpoints 500 with "config.toJSON is not a function" (v2.4 + develop)
## Summary

`GET /api/ssoConfig?provider=AzureAD` and `PUT /api/ssoConfig?provider=AzureAD`
both return **HTTP 500** with body:

```json
{"message":"Internal Server Error","error":"config.toJSON is not a function"}
```

This makes the **Settings → SSO** tab unusable: you cannot save an Entra ID
configuration, and once a partial row exists the tab 500s on load.

## Root cause

`Servers/utils/ssoConfig.utils.ts` — `getSSOConfigQuery()` (and
`saveSSOConfigQuery()`) fetch rows with a **raw** `sequelize.query(...)`:

```ts
export const getSSOConfigQuery = async (...): Promise<SSOConfigurationModel | undefined> => {
  const result = (await sequelize.query(
    `SELECT * FROM sso_configurations ...`,
    ...
  )) as [SSOConfigurationModel[], number];   // <-- cast is misleading
  return result[0][0];                        // <-- this is a plain POJO, NOT a model instance
};
```

The `as [SSOConfigurationModel[], number]` cast is incorrect: a raw
`sequelize.query` returns plain objects, which do **not** have Sequelize instance
methods like `.toJSON()`.

`Servers/controllers/ssoConfig.ctrl.ts` then calls `.toJSON()` on that POJO:

```ts
const config = await getSSOConfigQuery(organizationId, provider);
if (!config) return res.status(404)...;
return res.status(200).json({
  ...config.toJSON(),                 // <-- TypeError: config.toJSON is not a function
  config_data: maskConfig(config.config_data),
});
```

Same pattern in `saveSSOConfig` (`...saved.toJSON()`). Because the INSERT in
`saveSSOConfigQuery` succeeds *before* the controller throws on `.toJSON()`, the
save appears to fail (500) but a row IS persisted — after which every subsequent
`GET` also 500s on the same `.toJSON()` call.

## Reproduction

1. Deploy v2.4, log in as an org Admin, open **Settings → SSO**.
2. First load: `GET /api/ssoConfig?provider=AzureAD` → 404 (fine, no config).
3. Fill tenant/client ID/secret, click Save → `PUT` → **500** (`config.toJSON is not a function`).
4. Reload the tab → `GET` now also **500** (the partial row persists and can't be serialized).

## Suggested fix

Either return a real model instance from the query, e.g.:

```ts
return SSOConfigurationModel.findOne({ where: { organization_id, provider } }) ?? undefined;
```

…or stop calling `.toJSON()` on the raw result in the controller (the object is
already a plain POJO, so just spread it directly):

```ts
return res.status(200).json({
  ...config,
  config_data: maskConfig(config.config_data),
});
```

The same fix applies to `getSSOConfig`, `saveSSOConfig`, and any other handler
that spreads `.toJSON()` over a `getSSOConfigQuery`/`saveSSOConfigQuery` result.

## Environment notes (our deployment)

Self-hosted on AWS ECS Fargate, v2.4. `SSO_ENABLED=true` and `SSO_SECRET` are set
correctly (`/api/ssoConfig/feature` returns `{"enabled":true}`), served over HTTPS.
The failure is purely in the SSO config read/save code path and is independent of
deployment/config.

xieyanran:
## Describe your changes

`getSSOConfigQuery`/`saveSSOConfigQuery` in `Servers/utils/ssoConfig.utils.ts` run raw `sequelize.query()` calls, which return plain objects at runtime, not `SSOConfigurationModel` instances (the `as [SSOConfigurationModel[], number]` cast is misleading — it only satisfies the compiler). `Servers/controllers/ssoConfig.ctrl.ts` then called `.toJSON()` on those plain objects, which doesn't exist there, so `GET`/`PUT /api/ssoConfig` both 500'd with `config.toJSON is not a function` / `saved.toJSON is not a function`. Since the `INSERT` in the save path succeeds before the controller throws, a partial row gets persisted and every subsequent `GET` 500s too — so the SSO settings tab is unusable end to end.

Fix: spread the result directly instead of calling `.toJSON()` on it, since it's already the right shape.

Added a regression test (`Servers/controllers/__tests__/ssoConfig.ctrl.test.ts`) with a plain-object mock that reproduces the crash on the old code and passes on the fix. I also ran this against a real local Postgres (fresh `sequelize db:migrate`, calling `saveSSOConfig`/`getSSOConfig` directly) to confirm the actual save → read round trip works and the secret gets masked correctly.

Fixes #4219

- [ ] I have labelled the PR correctly. (no label permissions as an external contributor)
- [ ] The issue I am working on is assigned to me. (commented on the issue to claim it, not yet assigned)
- [x] N/A — no UI/theme changes, backend only.
- [x] N/A — no UI changes.
- [x] N/A — no API shape/route change, existing endpoint behavior restored.
- [x] N/A — no new endpoint.
- [x] N/A — no route changes.
- [x] N/A — no new/modified organization-scoped table.

### DIFF
--- a/Servers/controllers/__tests__/ssoConfig.ctrl.test.ts
+++ b/Servers/controllers/__tests__/ssoConfig.ctrl.test.ts
@@ -1,0 +1,124 @@
+import { describe, it, expect, jest, beforeEach, afterEach } from "@jest/globals";
+import { Request, Response } from "express";
+
+jest.mock("../../utils/ssoConfig.utils", () => ({
+  isSSOFeatureEnabled: jest.fn(),
+  getSSOConfigQuery: jest.fn(),
+  getFirstEnabledSSOConfigQuery: jest.fn(),
+  getSSOCapableOrganizationsQuery: jest.fn(),
+  saveSSOConfigQuery: jest.fn(),
+  setSSOEnabledQuery: jest.fn(),
+}));
+jest.mock("../../utils/logger/fileLogger", () => ({
+  __esModule: true,
+  default: { error: jest.fn() },
+  logStructured: jest.fn(),
+}));
+jest.mock("../../utils/statusCode.utils", () => ({
+  STATUS_CODE: {
+    200: (data: any) => ({ message: "OK", data }),
+    201: (data: any) => ({ message: "Created", data }),
+    400: (data: any) => ({ message: "Bad Request", data }),
+    404: (data: any) => ({ message: "Not Found", data }),
+    500: (data: any) => ({ message: "Internal Server Error", data }),
+  },
+}));
+
+import { getSSOConfig, saveSSOConfig } from "../ssoConfig.ctrl";
+import { getSSOConfigQuery, saveSSOConfigQuery } from "../../utils/ssoConfig.utils";
+
+const mockGetConfig = getSSOConfigQuery as jest.MockedFunction<typeof getSSOConfigQuery>;
+const mockSaveConfig = saveSSOConfigQuery as jest.MockedFunction<typeof saveSSOConfigQuery>;
+
+function createReq(overrides?: Partial<Request>): Request {
+  return {
+    organizationId: 1,
+    query: { provider: "AzureAD" },
+    body: {},
+    ...overrides,
+  } as any;
+}
+
+function createRes(): any {
+  const res: any = {};
+  res.status = jest.fn<any>().mockReturnValue(res);
+  res.json = jest.fn<any>().mockReturnValue(res);
+  return res;
+}
+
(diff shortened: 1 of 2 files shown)


MuhammadKhalilzadeh:
@HarshP4585 and @Br0wnHammer , would you please review this PR once you get the time?

HarshP4585 (approved):

Br0wnHammer (approved):
LGTM!

MuhammadKhalilzadeh (approved):

